PR #26, opened 2015-01-06
Add BDT training and reading functionality

chrisburr:
See [readme](https://github.com/alexpearce/CharmProduction/blob/analysis-bdt/analysis/README.md).

This pull request adds the requirement of a [set_data_vars](https://github.com/alexpearce/CharmProduction/blob/analysis-bdt/analysis/charmproduction/ntuples/D0Tohh.py#L59) method to all ntuples classes. The idea is that this contains the location of the DaVinci ntuples so we can access the data for any mode by just running `mode.setup_data()`. Tomorrow I'll modify the AFS permissions so you can access the D0 DVntuples if you want them.


chrisburr:
I will also fix the Travis CI errors.

alexpearce:
Could you [`log`](https://github.com/alexpearce/CharmProduction/blob/2f9d25c6545127b37b314b3dbc5f3beb226f7f0f/analysis/charmproduction/__init__.py#L5) instead of `print`? Personally it makes the log messages stand out more, and helps differentiate information from warnings (for grep'ing and stuff). I usually do something like

```python
import logging as log

def stuff():
    # ...
    log.info('Helpful information')
    log.warning('Something looks odd')
    log.error('Computer on fire')
```

chrisburr:
Done

alexpearce:
Could you update the documentation in [`CharmProductionNtuple`](https://github.com/alexpearce/CharmProduction/blob/9a5aae17212dd3141652ec1d2864b19a83c30a5b/analysis/charmproduction/ntuples/CharmProductionNtuple.py#L9)? There are some more methods child classes should implement. Also could you please add docstrings to `load`, `setup_data`, and `change_selection` in the same class?

chrisburr:
Do you have access to AFS setup on your mac? If so is it in /afs/cern.ch/? Assuming you do you can test if the `--silent` flag fixes the clang issue by running (from the analysis folder):
```python
from charmproduction import ntuples, BDT
mode = ntuples.D0Tohh('MagBoth', 2011)
mode.setup_data()
BDT.train(mode)
BDT.read_all(mode)
BDT.evaluate(mode)
```

alexpearce:
Yes I do, and yes it is. I get

```
[                    ] 0%ERROR: Error: 127
```

chrisburr:
Wonderful, I assume you have ran `make` in BDTTools and have read access to `/afs/cern.ch/work/c/cburr/Data/MC/2011_MagDown_D0Tohh/1/DVntuple.root`?

alexpearce:
I do have access to that file, but I hadn't run `make` *cough*. When I do, I get

```
ld: unknown option: --no-as-needed
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

When I remove all instances of `-Wl,--no-as-needed` from the Makefile, everything compiles.

On running your snippet again, I now get

```
[                    ] 0%ERROR: Error: -11
```

One note: I had to create the `lib`, `bin`, and `obj` directories. We can create a `.gitkeep` file in each as placeholders, so the folders are created when the repo is cloned, or have the Makefile ensure those folders exist.

chrisburr:
`.gitkeep` I learn something new everyday :+1:

@dominikmuller I guess this is on the C++ side so over to you.

alexpearce:
If I do e.g. `./bin/TreeSplitter` with or without the `--silent` flag I still get a segmentation fault.

Issue #33, opened 2015-01-09
Save likelihood profile plots for all fit variables


alexpearce:
Inspecting [likelihood profiles](https://root.cern.ch/root/html600/RooAbsReal.html#RooAbsReal:createProfile) can tell you quickly whether you hit a local minimum during fitting, whether your parameter range is sensible, and other goodies.

It would useful to safe the likelihood of all fit variables, for fits to data and Monte Carlo, so that they can be inspected.

Issue #32, opened 2015-01-08
Monte Carlo IP χ2 fits not simultaneous


alexpearce:
The 2D fits in mass and IP χ<sup>2</sup> can [now](https://github.com/alexpearce/CharmProduction/tree/7ad6aaf22d595f70d2aff801acd100956cbeaf15) be done simultaneously in bins of transverse momentum and pseudorapidity. However, the initial fits to Monte Carlo data for the prompt and secondary signal species and to data sidebands for background species are done over a single, integrated pT-y bin.

This is a problem because if a parameter is to be split across pT-y bins, its value will be fixed to the single value obtained from the MC fits. Instead, the MC fits should provide a value for each bin. (In fact, it might not even be fixed, as when the new `RooRealVar` objects are created for each bin, I don't know if the constant property value is carried over from the 'master' variable.)

So, the MC fits need to be performed in the same binning as the final fit to data, and the values need to be carried across from the MC fits to the final fit. Part of #18 .

alexpearce:
As above in [commit `cb506a4`](https://github.com/alexpearce/CharmProduction/commit/cb506a46f4b95f460bbaef74264b64f385259b12), this issue will be closed once the `fitting` branch is merged with master.

I have confirmed that if a parameter is split between pT-y bins in an IP χ<sup>2</sup> prefit, then fixed, its value and uncertainty correctly carry over to the fit to data.

For example, if a parameter has a fitted value of 1.14 ± 0.05 in the prefit to data sidebands in the second pT and third y bin, and is required to be [fixed by the `Ntuple`](https://github.com/alexpearce/CharmProduction/blob/cb506a46f4b95f460bbaef74264b64f385259b12/analysis/charmproduction/ntuples.py#L114), that same parameter in the same bin after the real fit will have a value of 1.14 ± 0.05.)

Issue #31, opened 2015-01-07
Simultaneous fitting for 3D fits


alexpearce:
The [current revision of the `fitting` branch](https://github.com/alexpearce/CharmProduction/tree/7ad6aaf22d595f70d2aff801acd100956cbeaf15) only works for 2D fits (in mass and IP χ<sup>2</sup>). It should be made compatible for D* modes as well. Part of #18 .

alexpearce:
Although [`d4b675`](https://github.com/alexpearce/CharmProduction/commit/d4b675d5fa70c67bcb74b79cc591ac9b77e0bc40) technically closes this issue, the 3D fits have trouble converging. In particular they seem to be very sensitive to the initial values of the various yields (there are two more species in 3D fits than in 2D). We can open another issue for that, though.

Issue #29, opened 2015-01-07
Allow per-mode parameter splitting across phase space bins


alexpearce:
Currently the parameters split across pT-y bins are [hardcoded](https://github.com/alexpearce/CharmProduction/blob/fitting/analysis/charmproduction/fitting/mass_ipchi2.py#L38), but modes should be allowed to specify what parameters should be split.

alexpearce:
Closed in [`316c36`](https://github.com/alexpearce/CharmProduction/commit/316c3673b2cca689fffb5f3661e9f578071872d2).

PR #34, opened 2015-01-09
Fitting

alexpearce:
Two months later...

I have a [few](https://github.com/alexpearce/CharmProduction/issues/18) [~~more~~](https://github.com/alexpearce/CharmProduction/issues/31) [~~things~~](https://github.com/alexpearce/CharmProduction/issues/32) I'd like to finish before merging, and [some](https://github.com/alexpearce/CharmProduction/issues/28) [other](https://github.com/alexpearce/CharmProduction/issues/29) [things](https://github.com/alexpearce/CharmProduction/issues/33) that can wait to have their own features branches, but the fitting framework is ready to be tested by everyone.

This won't be a completely clean merge as [`ntuples.py`](https://github.com/alexpearce/CharmProduction/blob/73f56e45281145fbeac4fd1a6064a4528762eb75/analysis/charmproduction/ntuples.py) as been split per-mode in [`ntuples/` on `analysis`](https://github.com/alexpearce/CharmProduction/tree/f0777a78d5278063fe796f7ee6a5ca7dce010328/analysis/charmproduction/ntuples), but it shouldn't be too much work.

## Testing

The [`fitting.py` script](https://github.com/alexpearce/CharmProduction/blob/73f56e45281145fbeac4fd1a6064a4528762eb75/analysis/fitting.py) is essentially the user guide on how to use the fitting framework. Currently, it can perform a 2D fit in mass and log<sub>10</sub>(IP χ<sup>2</sup>) simultaneously in pT-y bins, saving a bunch of plots in `analysis/output/`. It should be able to to the same in 3D in a couple of upcoming commits. Paths to ntuples are sort-of hardcoded.

What would be useful would be for you to test the fitting on your modes, using 2011 collision data and Monte Carlo. The MC data can be cheated or stripped, maybe the most useful is fully selected but I'd need to think about that. If there's something that's not clear or doesn't work how you'd expect, I'll consider it a bug.


alexpearce:
I have gone and *slightly* screwed up by merging `master` in to `fitting`, so merging with `analysis` will effectively merge `master` in to `analysis`.

…

chrisburr:
That needed to be done so :+1:

alexpearce:
I think this is ready to merge. If there aren't any major comments I'll merge tomorrow morning.

chrisburr:
I think it would be nice if `ntuples.py` was split up to make merging `analysis-bdt` easier.

alexpearce:
But I guess once during the merge, I'll get merge conflicts because of that? So I can split up `ntuples.py` then? Or do you think it would just be easier/cleaner to make a commit with the splitting before the merge?

chrisburr:
I didn't notice this isn't already in a merged state with `analysis`. I'm not sure, it depends if it will be easier to merge `analysis-bdt`+`analysis-without-magic` into `fitting` or vise versa.

PR #36, opened 2015-01-14
Analysis efficiency

chrisburr:
While this isn't polished I feel it is complete enough to merge to keep the branches consistent. Cheated and Generator MC is now downloaded in `setup_data` and `CharmProductionNtuples` can implement a post bdt selection.


analysis/charmproduction/efficiencies/plotting.py
```diff
@@ -0,0 +1,40 @@
...
+from utilities import get_real, parse_formula
+
+
+def binned_mass_plot(mode, bdt_cut, bdt_var='BDT_RESULT'):
+    # Get the data
+    ntuple = get_real(mode, bdt_var)
+    # Make the array of TH1Fs
+    mass_var, mass_low, mass_high = mode.mass_var_info()
+    histos = []
+    for x in range(5):
```
alexpearce: This looks like it's assuming a particular pT-y binning, which I expect will be different for each mode. I guess you might be waiting for an implementation of per-mode binning schemas from me, and if so it will look something [like this](https://github.com/alexpearce/CharmProduction/blob/a911866e2e022b879438c4571b596ef46c8dccb6/analysis/charmproduction/ntuples.py#L82-L89).

analysis/charmproduction/efficiencies/plotting.py
```diff
@@ -0,0 +1,40 @@
...
+        temp_histos = []
+        for y in range(8):
+            name = '{0}_{1}'.format(x, y)
+            temp_histos.append(ROOT.TH1F(name, name, 30, mass_low, mass_high))
+        histos.append(temp_histos)
+    # Make a lambda to check the post bdt cuts
+    formula = parse_formula(mode.postselection)
+    # Event loop
+    for e in ntuple:
```
alexpearce: Best to compute the various variable names outside the loop.

analysis/charmproduction/efficiencies/plotting.py
```diff
@@ -0,0 +1,40 @@
...
+        y = int(2*(ntuple.val('{0}_Y'.format(mode.mother_branch))-2))
+        pt = int(ntuple.val('{0}_PT'.format(mode.mother_branch))/1000)
+        if ntuple.val(bdt_var) > bdt_cut and formula(ntuple):
+            # Bin the event if it is within our phase-space
+            try:
+                mass = ntuple.val('{0}_MM'.format(mode.mother_branch))
```
alexpearce: Maybe move this line outside the `try`, as that's not what you're testing?

analysis/charmproduction/efficiencies/plotting.py
alexpearce: Maybe for now it would be nice to have `NBINS_Y` and `NBINS_PT` (or whatever) as constants in this file, and use those throughout rather than the magic numbers.

chrisburr: Can I still assume it will still be in the the range of 2->4.5 for rapidity and 0->8000MeV for transverse momentum?

analysis/charmproduction/efficiencies/selection.py
```diff
@@ -0,0 +1,29 @@
...
+from utilities import get_stripped, draw
+
+
+def selection(mode, bdt_cut, bdt_var='BDT_RESULT'):
+    ntuple = get_stripped(mode, bdt_var)
+    nstripped = numpy.zeros((5, 8))
+    nbdt = numpy.zeros((5, 8))
+    dropped = 0
+    # Loop over all candidate events
+    for e in ntuple:
```
alexpearce: As above moving string manipulation outside the loop.

analysis/charmproduction/efficiencies/plotting.py
chrisburr: Done

analysis/charmproduction/efficiencies/plotting.py
chrisburr: Done

analysis/charmproduction/efficiencies/plotting.py
alexpearce: Maybe not, you could ask [`transverse_momentum_var_info`](https://github.com/alexpearce/CharmProduction/blob/a911866e2e022b879438c4571b596ef46c8dccb6/analysis/charmproduction/ntuples.py#L74) and [`rapidity_var_info`](https://github.com/alexpearce/CharmProduction/blob/a911866e2e022b879438c4571b596ef46c8dccb6/analysis/charmproduction/ntuples.py#L78).

analysis/charmproduction/efficiencies/selection.py
chrisburr: Done

analysis/charmproduction/efficiencies/utilities.py
```diff
@@ -0,0 +1,105 @@
...
+        pt = int(ntuple.val(pt_name.format(mother_branch))/1000)
+        try:
+            result[y, pt] += 1
+        except IndexError:
+            dropped += 1
+    log.info("Dropped {0} out of range events".format(dropped))
+    return result
+
+
+def draw(before, after, mode, title):
```
alexpearce: Might belong better in `efficiencies/plotting.py`.

chrisburr: I don't see why? The comment in inaccurate, it doesn't make any plots it just loops over `ntuple` counting how many events fit in each bin and then returns a 2d array of integers.

analysis/charmproduction/ntuples/DstToD0pi_D0Tohh.py
```diff
@@ -1,22 +1,17 @@
 from charmproduction import config
 
-from CharmProductionNtuple import CharmProductionNtuple
+from D0Tohh import D0Tohh
 
 
-class DstToD0pi_D0Tohh(CharmProductionNtuple):
+class DstToD0pi_D0Tohh(D0Tohh):
```
alexpearce: I'm not sure if inheriting from `D0Tohh` really makes sense. I mean, maybe at the moment there are some methods that the tagged decay shares with the untagged, but is there really an inheritance relationship?

analysis/charmproduction/efficiencies/utilities.py
alexpearce: It creates a histogram and saves it. (I'm commenting on the `draw` method)

alexpearce:
Looks nice! A few general comments.

1. As we'll want to run over different datasets, you'll probably want to accept `polarity` and `year` as well as `mode` in some of the methods here.
2. For consistency, `import` from `charmproduction.efficiencies.<whatever>` (I don't feel too strongly about this though)
3. A lot of magic strings would be good candidates for going in `charmproduction.config` like `"DVntuple_REAL"` and `"BDT_RESULT"`, and things like `"Tuple{0}/DecayTree"` could go there too.

analysis/charmproduction/efficiencies/utilities.py
chrisburr: Sorry I was reading up rather than down. Fixed.

analysis/charmproduction/efficiencies/plotting.py
chrisburr: I've made a method in utilities which provides these constants so when fitting is merged it can be easily modified.

alexpearce: Ah yes I forgot those methods were introduced in `fitting`. Sounds like a good step in the meantime.

chrisburr:
1. This information is effectively taken from `mode.output_path` and should never point to data from another year/mode/polarity.
2. I agree, mainly as ROOT doesn't work with python 2 on my laptop and python 3 doesn't like the current imports.
3. Agreed.

alexpearce:
OK, all fine by me. I'm happy to make this Python 3 compatible, although I don't know PyROOT's compatibility, so we can open an issue/PR to work on it if you like.

chrisburr:
All the comments should be implemented except polarity splitting, moving strings to `charmproduction.config` and absolute imports all of which need to be changed in analysis as a whole. Is everyone happy for me to merge this?

alexpearce:
:+1:

PR #38, opened 2015-01-15
Split polarities correctly

chrisburr:
* Polarities are no longer combined for anything other than BDT training.
* The `set_data_vars` method has been changed to be cleaner, see [D0Tohh](https://github.com/alexpearce/CharmProduction/blob/analysis-without-magic/analysis/charmproduction/ntuples/D0Tohh.py#L66).
* Many of the hard coded strings have been moved to `charmproduction.config`


analysis/charmproduction/ntuples/get_data.py
```diff
@@ -91,16 +106,15 @@ def download(mode, file_name, down_path, down_start, down_end,
         time.sleep(0.05)
 
     # Merge the output
-    output = '{0}/{1}'.format(mode.output_path, file_name)
-    merge(output, '{0}_*'.format(output), True)
+    merge(filename, '{0}_*'.format(filename), delete=True)
 
 
-def merge(dest, source, delete):
+def merge(dest, source, delete=False, source2=''):
```
alexpearce: Why `source` and `source2`?

analysis/charmproduction/utilities.py
```diff
@@ -53,3 +53,16 @@ def ensure_directory_exists(path):
     if not directory_exists(path):
         log.info('Creating directory structure {0}'.format(path))
         os.makedirs(path)
+
+
+def recursive_dict_init(keys_list, d=dict(), previous_keys=[]):
```
alexpearce: So here's something that bit me just the other day! Don't use mutable objects as default arguments, because they are initialised once at the function definition rather than invocation. Try this:

```python
import random
import string

def f(d=dict()):
    d[''.join(random.choice(string.ascii_lowercase) for _ in range(5))] = 'wat'
    return d

for _ in range(3):
    print f()
```

You'll see a single dictionary printed with three keys, rather than a dictionary with one (different) key each time. The same goes for lists.

I [used `None` and created the mutable object in the function body](https://github.com/alexpearce/CharmProduction/blob/15c846f03e89f33587b848d990c7152d4e1efe04/analysis/charmproduction/containers.py#L3-L17) instead.

analysis/charmproduction/ntuples/get_data.py
chrisburr: `source2` is needed for [merging the two polarities](https://github.com/alexpearce/CharmProduction/blob/analysis-without-magic/analysis/charmproduction/ntuples/CharmProductionNtuple.py#L181) for the BDT training data. It is also used when merging the minbias and MC to make "real" data for 2015 (previously this was done with a crude use of wildcards).

analysis/charmproduction/utilities.py
chrisburr: I should have seen that coming, `previous_keys=[]` is also unsafe then?

analysis/charmproduction/ntuples/get_data.py
alexpearce: OK, I guess you could get the same effect by joining `source` and `source2` with a space in between the strings then?

In that case a better API might be either to accept a list of `sources` and then pass `' '.join(sources)` to `hadd` and `rm`.

(This will change things where you have method calls only with `source`, i.e. you'll have to wrap the argument in `[]`, so I'll leave this to your judgement to decide whether an API change is worth it.)

analysis/charmproduction/BDT/Read.py
```diff
@@ -3,44 +3,45 @@
...
     # Input files
     command.append('--files')
-    command.append('{0}/{1}'.format(ntuple.output_path, input_file))
+    command.append(input_file)
 
     # Output files
-    output_file = '{0}_{1}.root'.format(input_file[:-5], bdt_var)
     command.append('--output-file')
-    command.append('{0}/{1}'.format(ntuple.output_path, output_file))
+    command.append('{0}_{1}.root'.format(input_file[:-5], bdt_var))
```
alexpearce: This `'{0}_{1}'` might be candidate to go in `config`, as you [already have the output filenames in there](https://github.com/alexpearce/CharmProduction/blob/analysis-without-magic/analysis/charmproduction/config.py#L58-L67).

chrisburr: I didn't in this case I didn't use the filename from `config` as this guarantees the newly made friend tree is next to the correct ROOT file if someone calls `read` directly. I could move `{0}_{1}.root` but I didn't think that adds anything unless you disagree?

alexpearce: Nope, that sounds fine :+1:

PR #39, opened 2015-01-15
Add analysis

chrisburr:
Analysis is now in a mature enough state to be merged into master.


alexpearce:
Epic. I will have a double check of things tomorrow morning, running a few things to check if the merges broke anything (man I should write some tests).

alexpearce:
I've only just seen that [`D0Tohh` defines an `add` method](https://github.com/alexpearce/CharmProduction/blob/5e4874dee8cbc4bde245dcb34257ab1046fa1a24/analysis/charmproduction/ntuples/D0Tohh.py#L63) which is overriding [`Ntuple.add`](https://github.com/alexpearce/CharmProduction/blob/5e4874dee8cbc4bde245dcb34257ab1046fa1a24/analysis/charmproduction/Ntuple.py#L74).

Seems to me like `D0Tohh.add` is adding stuff to `self.years` which is used in [`CharmProductionNtuple. get_remote_data_paths`](https://github.com/alexpearce/CharmProduction/blob/5e4874dee8cbc4bde245dcb34257ab1046fa1a24/analysis/charmproduction/ntuples/CharmProductionNtuple.py#L105), so how about `D0Tohh.add_remote_data_path` instead?

chrisburr:
Sorry it was supposed to be moved into `CharmProductionNtuple`.

alexpearce:
do it :+1:

Issue #37, opened 2015-01-14
CharmProductionNtuple is messy


alexpearce:
I don't like the look of all the [`var_info`](https://github.com/alexpearce/CharmProduction/blob/15c846f03e89f33587b848d990c7152d4e1efe04/analysis/charmproduction/ntuples.py#L76) and related methods (which were introduced by me, sorry!). I propose to replace them with something more along the lines of this.

```python
class CharmProductionNtuple(Ntuple.Ntuple):
    # ...
    def __init__(self, polarity, year):
        # ...
        self.mass_var = containers.Variable(...)
        # More variables here...
    # ...
```

with a new [`container` class](https://github.com/alexpearce/CharmProduction/blob/395773ce20f6ab6e6a9643ce7279cc1f77653f69/analysis/charmproduction/containers.py) called `Variable` (better names welcome!)

```python
class Variable(object):
    def __init__(self, name, title, binning, unit=''):
        """Initialise a new Variable object.

        Keyword arguments:
        name -- Branch name this Variable represents
        title -- Pretty name to display on plots etc. in TLatex format
        binning -- List of bin edges including lowest and highest
                        e.g. binning=[1, 3, 4, 5, 10]
        unit -- Physical unit as a string in TLatex format (default: '')
        """
        if sorted(binning) != binning:
            # Raise an exception or whatever, user should pass ordered binning
        self.name = name
        self.title = title
        self.binning = ROOT.RooBinning(binning[0], binning[-1])
        for b in binning[1:-2]:
            self.binning.addBoundary(b)
        self.unit = unit
````

Then rather than having [very ugly code](https://github.com/alexpearce/CharmProduction/blob/15c846f03e89f33587b848d990c7152d4e1efe04/analysis/charmproduction/fitting/utilities.py#L42-L48) expanding the [current tuples of information](https://github.com/alexpearce/CharmProduction/blob/15c846f03e89f33587b848d990c7152d4e1efe04/analysis/charmproduction/ntuples.py#L76), we can just do `ntuple.mass_var.name` or whatever.

Thoughts? I'll work on this once `fitting` is merged with `analysis`, as code outside `fitting` is using the `var_info` methods too.

chrisburr:
This looks nice. As an extension we could have a `variables` module with a file for each variable we use so we can avoid things like defining the LaTeX names repeatedly.

alexpearce:
Closed with merging #40.

PR #40, opened 2015-01-16
Variable container

alexpearce:
Addresses #37 by implementing the `Variable` object in `containers`.

Cleans up quite a bit of code and I think makes stuff generally tidier and make more sense. It should also make adding extra variables much simpler.


analysis/charmproduction/efficiencies/utilities.py
```diff
@@ -75,7 +75,7 @@ def formula_funct(x):
     return formula_funct
 
 
-def count_binned(ntuple, mode, pt_name='{0}_PT'):
+def count_binned(ntuple, mode):
```
alexpearce: @chrisburr I deleted `pt_name` as I couldn't see it used.

chrisburr: It is used when counting the cheated tuples [here](https://github.com/alexpearce/CharmProduction/blob/variable-container/analysis/charmproduction/efficiencies/reconstruction.py#L12).

alexpearce: Ah right yes sorry, I thought I couldn't see it used in the method itself but it was right there… I'll add in back.

It would be nicer if these sorts of changes could be encapsulated in the `Ntuple`, like if there was a `CharmProductionGeneratorNtuple` or something that inherited from `CharmProductionNtuple` and changed a few variable values, but then you'd have to have corresponding `D0TohhGenerator` and everything…

alexpearce: Do you think something like this would be better in the meantime?

```python
def count_binned(ntuple, mode, pt_name=None):
    if not pt_name:
        pt_name = mode.transverse_momentum_var.name
    # ...
```

So that we can use the name already in the ntuple.

alexpearce:
I'm undecided as to whether the `_var` suffix just makes everything messier without being clearer.

```python
# Current
ntuple.mass_var.name
ntuple.transverse_momentum_var.binning

# Proposed?
ntuple.mass.name
ntuple.transverse_momentum.binning
```

analysis/charmproduction/efficiencies/utilities.py
alexpearce: I have [reimplemented `pt_name` in `0e2c94`](https://github.com/alexpearce/CharmProduction/commit/0e2c9420758210f38068c2263857d82f81233123) as above.

Issue #41, opened 2015-01-16
TTreeFormula


alexpearce:
I think the way @chrisburr has [parsed TTree cut strings](https://github.com/alexpearce/CharmProduction/blob/master/analysis/charmproduction/efficiencies/plotting.py#L80) [in Python lambdas](https://github.com/alexpearce/CharmProduction/blob/master/analysis/charmproduction/efficiencies/utilities.py#L59) is pretty darn cool. But, if there's a built-in way should we use that instead? There's an example of using [`TTreeFormula`](https://root.cern.ch/root/html600/TTreeFormula) on the [ROOT forums](https://root.cern.ch/phpBB3/viewtopic.php?f=14&t=16775#p78460), which I'll reproduce here.

```python
import ROOT

input_tree = # ...
formula = ROOT.TTreeFormula('name', 'expression_using_leaves_of_the_tree', input_tree)
formula.GetNdata()
for event in input_tree:
    val = formula.EvalInstance()
```

Which is neat. Thoughts?

PR #43, opened 2015-01-19
Replace custom parse_formula with TTreeFormula.

alexpearce:
This is a example implementation of replacing [`parse_formula`](https://github.com/alexpearce/CharmProduction/blob/8f6beb193c21867798fe27d39bbde8e43214a97d/analysis/charmproduction/efficiencies/utilities.py#L59) with ROOT's [`TTreeFormula`](https://root.cern.ch/root/html600/TTreeFormula), see #41. Check out [this gist](https://gist.github.com/alexpearce/4b45f553ecae48015913) I used as a quick test of `TTreeFormula`.

@chrisburr I've made this seeing the code at first glance, there might be something I missed that `parse_formula` is a better fit for than `TTreeFormula`. Please do feel free to reject!


dominikmuller:
No, but it returns a double. And sometimes I got some E-230 numbers instead of 0 for stuff that is false.

alexpearce:
@chrisburr Merge?

PR #44, opened 2015-01-19
Add unit tests

alexpearce:
I've modified the [Travis CI configuration](https://github.com/alexpearce/CharmProduction/blob/aa2c912622f22c75b8ddd4fb6dc579c332de649f/.travis.yml) to install ROOT, so now we can start writing proper unit tests.

This PR will live for a little while, but hopefully no longer than a week. I'll upload some “dummy” data so that we have ntuples to test with, and will write some tests to cover things that are working now.

The idea here to to make [large pull requests](https://github.com/alexpearce/CharmProduction/pull/39) and [code restructuring](https://github.com/alexpearce/CharmProduction/pull/40) less scary, requiring less manual testing by ourselves to check that things still work.


chrisburr:
This is going to slow down our "builds" a bit! I like the idea though. Where were you thinking of storing the dummy data?

alexpearce:
I was thinking of just uploading one or two TTrees from the 2010 dataset. They technically won't contain anything that hasn't been published (only really need a few branches). As for the actual ‘where’, probably just in a public dropbox folder.

alexpearce:
I'd rather just get what there is now merged so I can keep writing tests for new features.

Any objections to all this? (Apart from the failing test which I will fix…)

chrisburr:
None at all, are you going to fix or removed the ROOT5 test?

alexpearce:
Good question. Any thoughts? I guess the Python module shouldn't really have any trouble supporting both, but if only want to support 6 we can drop it.

chrisburr:
I don't see a reason for keeping support for 5 and if we hit issues I suspect we'll just drop it in any case.

PR #46, opened 2015-01-26
Likelihood profiles

alexpearce:
Likelihood scans let you know if you parameter range is sensible, and whether the fit had trouble finding a minimum.

likelihood_mu_m_sig
likelihood_sigmal_ip_sec

The top scan looks nice, whilst the bottom looks, to use the technical term, ‘a little funky’.

This PR saves likelihood scans for every simultaneous PDF parameter in the output of the [`fitting.py`](https://github.com/alexpearce/CharmProduction/blob/475bc4347124ef47849f97fb8cfa2f21e8f40f78/analysis/fitting.py) script.


dominikmuller:
I like those plots, the top one looks really good. I am not sure about the bottom one. The shape of the log likelihood only has to be quadratic around the minimum if you do the taylor expansion of the likelihood. And for the uncertainties you go 1/2 on the y-axis so maybe it does not look that bad after all once you zoom in,
You integrate over all other parameters to get these curves right? I could imagine some funny results if one of the other parameters is weirdly correlated and thus distorts the plotted variable in a projection.

alexpearce:
Sounds like you know more about these than I do, very good :wink:

I wasn't sure about the y-axis scale, when you say go 1/2 do you mean go use [0, 0.5] or just half the current scale?

I'm pretty sure all other parameters are integrated over, though I will double check the RooFit code. There are only a couple of plots that look like the bottom one. These were made before tweaking any parameter ranges (which is why I made the plots) so I wonder if they will improve after that, like the floated Gaussian widths are technically allowed to go to zero, but physically they won't go that low.

dominikmuller:
The uncertainty is determined by computing the which solves log L(min) - log L(min+sigma) = -0.5
So I would plot it in a range of maybe [f(min),f(min)+4]

alexpearce:
I've copied two output files into `/afs/cern.ch/user/a/apearce/public/forDominik/likelihoodScans`. The `SmallRange` one has the y-axis range as [0, 4] whilst the `LargeRange` has it as [0, `nll_curve.GetMaximum()`].

The `SmallRange` ones are very sharp, does this look how you would expect?

dominikmuller:
The uncertainty on the fitted parameter should match the width at log L = 0.5 divided by 2. So if that's the case all is fine.

alexpearce:
I've updated the plots to show the LL curve ±5σ around the fitted value, and zoomed in the y-axis as you suggested @dominikmuller.

mu_m_sig
sigmal_ip_sec


 Things like `mu_m_sig` look nice as expected, `sigmaL_ip_sec` still looks funky. Comparing the zoomed in plot with the one in the [first post](https://github.com/alexpearce/CharmProduction/pull/46#issue-55496665), looks like the fitter is having trouble with this variable (there are a couple of others that look like this).

alexpearce:
Any comments on the code? If not I'll merge this afternoon.

dominikmuller:
Looks good. Did you got around to try profile likelihoods?

chrisburr:
:+1:

alexpearce:
@dominikmuller I tried profile likelihoods but didn't even get to look at the plots because it took _ages_. But you're right, I should at least investigate it once before merging, to see what changes.

dominikmuller:
I know. It scans the parameter and at each step it maximises wrt all other parameters

alexpearce:
This is good to go.

I tried profile likelihood but a) it takes ages and b) I didn't really understand the results, which in general look weirder. I've added likelihood scans shown across a wider range, as well as ± 5σ, for ‘debugging’ if a particular parameter looks odd.

chrisburr:
I'm happy for it to be merged.

dominikmuller:
Would be interested to see the profile likelihood plots you got. But nonetheless, go for merge!

Issue #50, opened 2015-01-28
Data source to test stripping cuts on


chrisburr:
We need to decide which data to test the stripping cuts on.

### Signal

I guess the best thing to use is Cheated MC. If so should we use:
 - the MC the BDT is trained on?
 - random sample from the MC set which is used for calculating efficiencies?
 - a separate sample of the MC?

My thinking is a separate MC sample seeming as we can always generate more MC if needed.

### Background
I think this is a little tricker. Any of the above options using minbias MC will ignore any combinatorial backgrounds which I assume we need? So is the only solution to have another tuple in the `DVntuple.root` file with modified stripping line to have no cuts applied at all?

To get a feel for the number of events in the full minbias dataset in both polarities we have:

Mode | Cheated BKGCAT < 11 | Cheated BKGCAT > 10 | Stripped BKGCAT < 11 | Stripped BKGCAT > 10
------------ | ------------- | ------------- | ------------- | -------------
D0Tohh | 19,412 | 138,112 | 5,251 | 1,123,538

alexpearce:
### Signal

Would have thought to use the cheated MC for training the preselection, then use preselected and cheated MC for training and testing the BDT.

### Background

An appropriate background sample for some modes might be minbias that has been combined in the same way as signal, e.g. for D0ToKpipipi use four track combinations.

chrisburr:
So the best option is probably to create another ntuple during `setup_data` called `train_stripping_sig.root`.

For the background minbias am I right in thinking we need to run DaVinci again to create this new tree?

alexpearce:
Hmm unfortunately I think so. It will basically be the cheated ntuples but without `mcMatch`.

chrisburr:
As we have decided this isn't needed for now I'm going to close this.

PR #51, opened 2015-01-29
Add prelimary stripping optimisation script

chrisburr:
This pull request won't be ready until #50 has been resolved however until then comments are welcome.

Three new methods are added:

`stripping_optimisation.calculate_cuts(mode, variables)`
 - Uses TMVA to do a rectangular cut optimisation of the variables listed in variables. Currently takes 1-2 hours with 7 variables.

`stripping_optimisation.find_optimal(test_mode, bkg_supression)`
 - Looks at the cut values in the weights file (produced by TMVA) to find the best cuts to reduce the background to bkg_supression while keeping the signal efficiency as high as possible.

`charmproduction.efficiencies.stripping_by_cut(mode, cuts)`
 - Find the signal efficiency of each variable in bins of pT and rapidity.
 - Also find the total efficiency of the cuts combined.


alexpearce:
Wow that is slow. How many events are you training with?

chrisburr:
Hadn't thought about that given its a "useless" dataset, turns out to be 1.1million

dominikmuller:
Very nice job!

alexpearce:
OK that's probably overkill for training! What's it like with a smaller number, say 50k?

chrisburr:
Using the D0 BDT training data it takes a minute or so

alexpearce:
That's more like it, makes it much nicer to iterate quickly. How many events is that?

So, does TMVA still give you the variables ranked by discriminating power, and the usual correlation matrix and stuff? What figure of metric does it use to find the optimal cut values?

My idea is to throw a whole bunch of variables in there, at least as many as in the current stripping, and see how everything does. Then throw out the crap ones (put in some extras as well to see how they do).

If TMVA spectator variables do what I think they do, we should put IP chisq in as one, to make sure we don't cut on anything strongly correlated with it.

chrisburr:
Yeah TMVA does the usual ranking and correlation matrix etc.

The algorithm outputs 100 bins that scale linearly with signal efficiency each specifying a set of cuts which minimises the background. Currently I have it set so we can specify a background efficiency and it will read the weights file and return the loosest cuts that meet the background requirement. On the plots below the y axis is efficiency and the x axis is related to which "bin" of cuts is used.

strippingcutsexample_bkg
strippingcutsexample_sig

The idea of throwing in everything is what I was hoping for as well. If spectator variables work like that that would be really useful, I'll look into them tomorrow.

dominikmuller:
Unfortunately they do not... I gave it a try locally adding spectator variables.
They are included in the training and testing tree which TMVA produces in its output rootfile but not in the correlation matrix or input variables distributions.

chrisburr:
In that case we can train on all variables just to get the matrix then do a proper training after.

dominikmuller:
That's for sure possible and probably not that difficult. I just checked TMVA source, it should be possible to modify TMVA to do it automatically.

alexpearce:
Oo neat. And a cut on the discriminant still applies rectangular cuts on the input variables?

Doing a dummy training to get the correlations sounds sensible. There are ways you can compute the correlation yourself, though if we can get it like that it's easy enough.

Looks good, would be interested to try this tomorrow.

chrisburr:
You don't cut on a discriminant exactly, you pick one of the 100 bins which will then give you a Yes/No response. The bin you chose actually contains a list of variables and cut values i.e. [('Min_D0_ETA', 2), ('Max_D0_ETA', 5), ('Min_D0_P', 100), ('Max_D0_P', 30000000)].

chrisburr:
Is it worth quickly tidying this up and merging it for if we want it at a later date? Or shall we just remove it?

alexpearce:
I think we should keep it, if only for the efficiency evaluation method, looks pretty useful.

chrisburr:
I can put the optimisation script in a gist if we don't want to keep it here else I think its ready to merge.

alexpearce:
Hmm let's keep it for, it's nicely self-contained and may come in useful.

PR #52, opened 2015-01-30
Unstripped ntuples

chrisburr:
This adds `UnstrippedTuple[MODE]` to the relevent MC ntuples for stripping selection studies. Test jobs are currently running and I hope to submit this to the grid at the end of the day barring any major issues.

I was also considering adding some of the new modes to this pull request?


chrisburr:
Note to myself: I think the pi0 should actually be using `StdLooseResolvedPi0` i.e. https://svnweb.cern.ch/trac/lhcb/browser/Stripping/trunk/Phys/StrippingArchive/python/StrippingArchive/Stripping20/StrippingDstarD2hhpi0.py#L14

alexpearce:
To make things more complicated, there are multiple containers of neutral pions because they can be reconstructed in different ways, similarly to how there are long-long and down-down Kshorts (depends where you reconstruct the decay vertex).

We always reconstruct neutral pions through the π⁰ → γγ decay channel, and the photons are detected by the calorimeters. Because the calo cells are of finite size, sometimes the two photons can be detected by the same cell (or cluster of cells); these are "merged" pions. Alternatively the photons can be detected separately, these are "resolved" pions, i.e. you were able to resolve two distinct photons. Generally, merged pions are higher momentum because the opening angle between the two photons is smaller.

[This presentation](https://indico.cern.ch/event/311278/contribution/2/material/slides/0.pdf) contains some more information, like why two types have different backgrounds and mass resolutions.

chrisburr:
I think this branch should be removed would you agree?

alexpearce:
I think so, considering the timings it seems like we'll have to find another background sample.

PR #53, opened 2015-01-31
Add stripping lines

alexpearce:
I'm including these in this repo so we can open issues on them and stuff. The changes in this branch haven't been included in [StrippingSelections](https://svnweb.cern.ch/trac/lhcb/browser/Stripping/trunk/Phys/StrippingSelections/python/StrippingSelections) yet, I'll wait until we have proto-lines for all modes. Which brings me too…

### Stripping lines for all modes

This is the list we need to complete to have stripping lines for everything we want.

- [x] [D0Tohh](https://github.com/alexpearce/CharmProduction/blob/1da82bbfbe4c436899272931a660688b33ca796c/stripping/StrippingD02HHForXSec.py)
- [x] [D0ToKpipipi](https://github.com/alexpearce/CharmProduction/blob/1da82bbfbe4c436899272931a660688b33ca796c/stripping/StrippingD02K3PiForXSec.py)
- [x] [DpToKpipi](https://github.com/alexpearce/CharmProduction/blob/1da82bbfbe4c436899272931a660688b33ca796c/stripping/StrippingD2HHHForXSec.py)
- [x] [DpToKKpi](https://github.com/alexpearce/CharmProduction/blob/1da82bbfbe4c436899272931a660688b33ca796c/stripping/StrippingD2HHHForXSec.py) (or [DpTophipi](https://github.com/alexpearce/CharmProduction/blob/1da82bbfbe4c436899272931a660688b33ca796c/stripping/StrippingD2PhiPiForXSec.py))
- [x] [DsTophipi](https://github.com/alexpearce/CharmProduction/blob/1da82bbfbe4c436899272931a660688b33ca796c/stripping/StrippingD2PhiPiForXSec.py)
- [x] [DsTopipipi](https://github.com/alexpearce/CharmProduction/blob/1da82bbfbe4c436899272931a660688b33ca796c/stripping/StrippingD2HHHForXSec.py)
- [x] [DstToD0pi_D0Tohh](https://github.com/alexpearce/CharmProduction/blob/1da82bbfbe4c436899272931a660688b33ca796c/stripping/StrippingD02HHForXSec.py)
- [x] [Dst0ToD0pi_D0ToKpipipi](https://github.com/alexpearce/CharmProduction/blob/1da82bbfbe4c436899272931a660688b33ca796c/stripping/StrippingD02K3PiForXSec.py)
- [ ] Dst0ToD0pi_D0Tohh
- [ ] DstToD0pi_D0ToKpipipi
- [x] [LcTopKpi](https://github.com/alexpearce/CharmProduction/blob/1da82bbfbe4c436899272931a660688b33ca796c/stripping/StrippingLambdac2PKPiForXSec.py)
- [x] [LcTopKK](https://github.com/alexpearce/CharmProduction/blob/1da82bbfbe4c436899272931a660688b33ca796c/stripping/StrippingLambdac2PKPiForXSec.py) (or [LcToppipi](https://github.com/alexpearce/CharmProduction/blob/1da82bbfbe4c436899272931a660688b33ca796c/stripping/StrippingLambdac2PKPiForXSec.py))
- [x] [ScToLcpi](https://github.com/alexpearce/CharmProduction/blob/1da82bbfbe4c436899272931a660688b33ca796c/stripping/StrippingLambdac2PKPiForXSec.py)
- [x] [Sc0ToLcpi](https://github.com/alexpearce/CharmProduction/blob/1da82bbfbe4c436899272931a660688b33ca796c/stripping/StrippingLambdac2PKPiForXSec.py)
- [ ] XicTopKpi
- [ ] XicTopKKpi

(Better suggestions for shorthand names?)

The D\*0 modes can go in the respective D0 lines, should be very similar selections to the D\*+ modes already in them. The Xi_c lines should have a new module. (We could capture the pKpi mode in the Lambda_c module by extending the mass window, but the mass difference is [quite](http://pdg8.lbl.gov/rpp2014v1/pdgLive/Particle.action?node=S033) [large](http://pdg8.lbl.gov/rpp2014v1/pdgLive/Particle.action?node=S045) so we'd probably get an unacceptable rate. Makes most sense to me to just duplicate the Lambda_c module and edit it.)

### Selections

I haven't altered the selections at all, other than changing [`ADOCAMAX`](https://twiki.cern.ch/twiki/bin/view/LHCb/LoKiArrayFunctions#LoKiAMAXDOCA) to [`ACUTDOCA`](https://twiki.cern.ch/twiki/bin/view/LHCb/LoKiHybridFilters#AMAXDOCA_AMINDOCA_Distance_of_cl) as apparently it's better for CPU usage. I've checked this by running each stripping line other 10k events before and after the alterations, then `diff`'ing the resulting two log files. I didn't see any changes in the number of selected candidates, only differences in timing (where fluctuations are expected), so concluded the modifications haven't changed the physics.

But still, we need to look at the current selections and update them. They are very consistent across the lines: the same sets of cuts, often with the same values. I think we should strive for something similar as it will make our lives easier.


alexpearce:
`Dst0ToD0pi_D0Tohh` and `Dst0ToD0pi_D0ToKpipipi` were added in [`f10b64`](https://github.com/alexpearce/CharmProduction/commit/f10b648945d95c3bd02ebd480b37f0994c3c7fde) and `XicTopKpi` and `XicTopKKpi` were added in [`66608f`](https://github.com/alexpearce/CharmProduction/commit/66608f50a6a5a1e9d96e709ac97e37565b4d4e7f).

chrisburr:
How do you feel about having a flag to disable the BDT entirely? Either as a separate variable in the configuration dictionary or by setting the BDT cut at something very loose like -1000.

alexpearce:
You not creating the BDT `Selection` object at all? Why disable it if the cut is at -1000 anyway, for a speed up?

chrisburr:
Partly for a speed up but mainly so the weights don't have to exist. I *think* it just ignores the BDT if the weights aren't there but it seems safer just not use it.

alexpearce:
Hmm, how about checking the key defining the weights file then?

```python
def makeMVASelection(self, name, inputSel):
    if not self.config['D0_MVA_Weights']:
        return inputSel
    # ...
```

chrisburr:
That works :+1:

alexpearce:
Done.

alexpearce:
I've added the stripping line for `DsstToDsgamma_DsTophipi` in [`StrippingD2PhiPiForXSec.py`](https://github.com/alexpearce/CharmProduction/blob/a7ad5b6559d3b0e103ecd5a42346514b941a2972/stripping/StrippingD2PhiPiForXSec.py).

It takes the (MVA selected) D+ built for the KKpi mode and adds a photon, making an upper delta mass cut and a loose vertex fit quality. I'm not certain that a “vertex” makes sense here, so maybe the fit quality cut should go… I'm also not sure of how tight the photon selection is. Should be tested on Dsst MC.

But anyway, that's the final line done! :tada:

PR #54, opened 2015-02-01
PIDcalib integration

dominikmuller:
Disclaimer: This is still wip, but maybe you have some better ideas on how to structure things.

To get it to work, you have to setup PIDcalib:

```bash
SetupUrania --build-env v2r4
getpack PIDCalib/PIDPerfTools v6r1
getpack PIDCalib/PIDPerfScripts v8r1
cd PIDCalib/PIDPerfScripts/cmt
cmt br cmt make
SetupUrania v2r4
```

The last line `SetupUrania v2r4` has to be repeated every time you want to run it as it sets necessary environment variables. To run any of the modules:

```
from charmproduction import efficiencies
```

The code consists of three steps. `efficiencies.pid.setup_calib(stripping)` creates the performance histograms for all cuts defined in `config.py` and all binnings defined in `binning.py`. 

```python
kaon_pid_vars= {
    'DLLKg8': 'DLLK > 8.0'
}
pion_pid_vars= {
    'DLLKl3': 'DLLK < 3.0'
}
pid_vars= {
        'K': kaon_pid_vars,
        'Pi': pion_pid_vars
}
```
The cuts can be subsequently accessed using their names and added to the modes.
I highly recommend to due so as the name of the cuts have to MATCH exactly: `'DLLK < 3.0'` is not the same as `'DLLK< 3.0'` and PIDcalib handles such input errors properly (`seg faults ...`). The calibration files are stored in a folder in my lxplus work directory, so ideally you won't have to run this part anyway.

The mode configuration has to be extended by matching the branch names with a track type and the PID cut:

```python
pid_cuts = [
    ('K','Dp_h1',config.pid_vars['K']['DLLKg8']),
    ('Pi','Dp_h2',config.pid_vars['Pi']['DLLKl3']),
    ('Pi','Dp_h3',config.pid_vars['Pi']['DLLKl3']),
]
```

I will add a utility function converting this into the old plain cut strings.
The second step `efficiencies.pid.ProcessSignalTree(mode)` creates a tree similar to the application of the BDT which can be befriended with the original tree and contains all PID weights. Lastly, `efficiencies.pid.CalculateStatErrorBinned(mode)` takes this tree, applies a binning in `PT` and `Y` and calculates the differential efficiency histogram for the applied cuts per binning scheme used.


dominikmuller:
And before I forget, the scripts are modified versions of the PIDcalib things and do not look as they will please travis. I'll fix all of that.

analysis/charmproduction/ntuples/DpToKpipi.py
```diff
@@ -58,6 +58,16 @@ class DpToKpipi(CharmProductionNtuple):
 
     postselection = ('(Dp_h1_ProbNNk>0.1)')
 
+    """Adding the PID cuts and the necessary information for PIDcalib.
+    The dictionaries are used to avoid mispelling of the cut names, which
+    have to match exactly with the ones in the performance histograms!!!
+    """
+    pid_cuts = [
+        ('K','Dp_h1',config.pid_vars['K']['DLLKg8']),
```
chrisburr: I think we should prioritise the ntuples files being nicer rather than making it easier to parse for the script to get PID efficiencies. We can have a PID cut described like `('K', 'Dp_h1', 'DLLK > 8.0')` then find all the cuts we're using with something like:
```python
cuts = []
for mode in config.modes:
    for pid_cut in mode.pid_cuts:
        cuts.append([pid_cut[0], pid_cut[2]])
# Get a unique list of all needed cuts
cuts = list(set(cuts))
```

chrisburr: Or maybe something more like this:
`config.species = [config.kaon, config.pion, config.proton]`
```python
cuts = dict()
for particle in config.species:
    cuts[particle] = []
for mode in config.modes:
    for particle, var_name, pid_cut in mode.pid_cuts:
        cuts[particle_type].append(pid_cut)
# Get a unique list of all needed cuts
for particle in config.species:
    cuts[particle] = list(set(cutss[particle]))
```

chrisburr: Or if we keep to using 'K' instead of config.kaon this could be a job for [autovivification](http://en.wikipedia.org/wiki/Autovivification#Python)! @alexpearce

analysis/charmproduction/ntuples/DpToKKpi.py
```diff
@@ -46,10 +46,13 @@ class DpToKKpi(CharmProductionNtuple):
...
-        'Dp_h1_ProbNNk>0.1',
-        'Dp_h2_ProbNNk>0.1'
-    )
+    pid_cuts = [
+        (config.kaon, 'Dp_h1', 'DLLK > 8.0'),
+        (config.kaon, 'Dp_h2', 'DLLK > 8.0'),
+        (config.pion, 'Dp_h3', 'DLLK < 3.0'),
+    ]
+
+    postselection = cut_strings.pid(pid_cuts)
```
dominikmuller: That's how the PID cuts are implemented for now.

In the creation of the performance histograms, it automatically finds all cuts applied and calculates the necessary histograms.

dominikmuller:
So, now that Travis is happy (thank you travis!) What about merging?

analysis/charmproduction/ntuples/CharmProductionNtuple.py
```diff
@@ -179,6 +182,23 @@ def setup_data(self):
...
+        def etastring(mode):
+            daughter_eta = '{0}_ETA>2.0&&{0}_ETA<5.0'
+            temp = []
+            for d in mode.head.all_daughters():
+                temp.append(daughter_eta.format(d.name))
+
+            return '&&'.join(temp)
+
+        def pstring(mode):
+            daughter_eta = '{0}_P>3200.&&{0}_P<100000.'
```
chrisburr: I presume this should be daughter p?

chrisburr: I would also prefer `p_string(mode)` and `eta_string(mode)` for consistency.

chrisburr: Also I thought 2000MeV was the lower bound for PID?

analysis/charmproduction/ntuples/CharmProductionNtuple.py
```diff
@@ -198,12 +218,28 @@ def setup_data(self):
                          .format(preselection, self.head.name))
 
             # Make the signal training data
-            get_data.download(self, config.local_bdt_sig, selection, True,
-                              split=True)
+            get_data.download(
+                self,
+                config.local_bdt_sig,
+                selection +
```
chrisburr: I would prefer this, and the second instance of it, on one line like:
`selection + '&&' + eta_string(self) + '&&' + p_string(self),`
or (as I slightly prefer):
`'&&'.join([selection, eta_string(self), p_string(self)]),`

analysis/charmproduction/ntuples/CharmProductionNtuple.py
```diff
@@ -179,6 +182,23 @@ def setup_data(self):
         path = config.eff_path.format(self.output_path, config.magdown, '')
         utilities.ensure_directory_exists(path)
 
+        # Cheating in a stripping cut on ETA to make pid calib behave.
```
chrisburr: I think this is the wrong place to be applying cuts but I guess the reason for doing this here is because of the event weights messing up when passing the tree to PIDCalib?

dominikmuller: It is only temporary as I expect the cuts to be in the stripping later on anyway. I wanted them to be applied such that they count towards the stripping selection which is determined from everything in the tuple vs the things in the cheated tuple.

analysis/charmproduction/ntuples/CharmProductionNtuple.py
dominikmuller: Will be changed.

analysis/charmproduction/ntuples/CharmProductionNtuple.py
dominikmuller: The lower bound for PIDcalib is whatever I choose in the binning. For now I picked the binning from the 2010 note. But I'll check if there are any recommendations when using the new calibration samples.

analysis/BDTTools/Makefile
```diff
@@ -28,19 +35,19 @@ Objects_TreeSplitter=$(patsubst %,$(ODIR)/%,$(_Objects_TreeSplitter))
...
 	@echo '-> ./bin/BDTControlPlots created!'
 
 TMVAClassificationApplication: $(Objects_TMVAClassificationApplication)
-	$(CC)  $(LDFLAGS) $(ROOTLIBS) -lTMVA $(Objects_TMVAClassificationApplication) -lm -lstdc++ -o ./bin/TMVAClassificationApplication
+	$(CC) -Wl,--no-as-needed  $(LDFLAGS) $(ROOTLIBS) -lTMVA $(Objects_TMVAClassificationApplication) -lm -o ./bin/TMVAClassificationApplication
 	@echo '-> ./bin/TMVAClassificationApplication created!'
 
 TreeSplitter: $(Objects_TreeSplitter)
-	$(CC)  $(LDFLAGS) $(ROOTLIBS) $(Objects_TreeSplitter) -lm -lstdc++ -o ./bin/TreeSplitter
+	$(CC) -Wl,--no-as-needed  $(LDFLAGS) $(ROOTLIBS) $(Objects_TreeSplitter) -lm -o ./bin/TreeSplitter
```
alexpearce: The `--no-as-needed` flag are not supported by Clang. Applying this patch gets things working for me.

```patch
diff --git a/analysis/BDTTools/Makefile b/analysis/BDTTools/Makefile
index c2f7681..a1768c0 100644
--- a/analysis/BDTTools/Makefile
+++ b/analysis/BDTTools/Makefile
@@ -39,15 +39,15 @@ TMVAClassification: $(Objects_TMVAClassification)
        @echo '-> ./bin/TMVAClassification created!'

 BDTControlPlots: $(Objects_BDTControlPlots)
-       $(CC) -Wl,--no-as-needed  $(LDFLAGS) -lTMVA $(ROOTLIBS) $(Objects_BDTControlPlots) -lm -o ./bin/BDTControlPlots
+       $(CC) -Wl $(LDFLAGS) -lTMVA $(ROOTLIBS) $(Objects_BDTControlPlots) -lm -o ./bin/BDTControlPlots
        @echo '-> ./bin/BDTControlPlots created!'

 TMVAClassificationApplication: $(Objects_TMVAClassificationApplication)
-       $(CC) -Wl,--no-as-needed  $(LDFLAGS) $(ROOTLIBS) -lTMVA $(Objects_TMVAClassificationApplication) -lm -o ./bin/TMVAClassificationAppli
+       $(CC) -Wl $(LDFLAGS) $(ROOTLIBS) -lTMVA $(Objects_TMVAClassificationApplication) -lm -o ./bin/TMVAClassificationApplication
        @echo '-> ./bin/TMVAClassificationApplication created!'

 TreeSplitter: $(Objects_TreeSplitter)
-       $(CC) -Wl,--no-as-needed  $(LDFLAGS) $(ROOTLIBS) $(Objects_TreeSplitter) -lm -o ./bin/TreeSplitter
+       $(CC) -Wl $(LDFLAGS) $(ROOTLIBS) $(Objects_TreeSplitter) -lm -o ./bin/TreeSplitter
        @echo '-> ./bin/TreeSplitter.x created!'

 dict/MyDict.cxx: include/ROOTex.h include/LinkDef.h
```

analysis/charmproduction/config.py
```diff
@@ -57,6 +57,15 @@
 cb_mc = '/afs/cern.ch/work/c/cburr/Data/MC/20'
 dm_mc = '/afs/cern.ch/work/d/dmuller/sample/Data/MC/20'
 
+# Storage path for PID track calibration samples
+calib_dir = '/afs/cern.ch/work/d/dmuller/CharmProduction/CalibrationFiles/'
```
alexpearce: So this is where the “performance histograms” created by the PIDCalib scripts will live, right? So you will run those scripts, put the files in this directory, and everything should work?

Also, I suppose having the same directory structure in this directory as in `output/` is probably a good idea.

chrisburr: Won't the structure in here be very different as we can use the same samples and binning for many modes so it makes more sense to be like:
```
protons
    p_0_1000_3000_5000
        eta_2_3_3.5_4_5
            some_sample.root
kaons
    p_0_1000_3000_5000
        eta_2_3_3.5_4_5
            some_sample.root
pions
    p_0_1000_3000_5000
        eta_2_3_3.5_4_5
            some_sample.root
```

analysis/charmproduction/efficiencies/utilities.py
```diff
@@ -130,3 +120,20 @@ def calculate_and_print(before, after, mode, title):
...
+    n_y_bins = mode.rapidity_var.binning.numBins()
+
+    histo = ROOT.TH2F(name, name, n_y_bins, y_arr, n_pT_bins, pT_arr)
+    histo.GetXaxis().SetTitle('#text{$y$}')
+    histo.GetYaxis().SetTitle('#text{$p_{T}$ [MeV/$c^2]$}')
+
+    histo.GetXaxis().CenterTitle()
+    histo.GetYaxis().CenterTitle()
+
+    return histo
```
alexpearce: I like this, really cleans things up!

We should use the stuff defined in the appropriate `Variable` instances for the LaTeX title and units.

```python
y_title = mode.rapidity_var.title
y_unit = mode.rapidity_var.unit
# ...
histo.GetXaxis().SetTitle(' '.join(y_title, y_unit]))
```

analysis/charmproduction/ntuples/CharmProductionNtuple.py
```diff
@@ -7,6 +7,7 @@
 
 
 class CharmProductionNtuple(Ntuple.Ntuple):
+
```
alexpearce: Delete this line.

analysis/charmproduction/ntuples/CharmProductionNtuple.py
alexpearce: Shouldn't these strings just be a part of `preselection` rather than having their own methods?

analysis/charmproduction/config.py
dominikmuller: I cannot split the files easily into different track types as that is done within PIDcalib and I would have to move the files around "manually" after their creation. I could split it into year ( well stripping actually), polarity and the employed binning scheme.

analysis/charmproduction/config.py
```diff
@@ -57,6 +57,15 @@
...
+
+# List of PID selection variables
+kaon = 'K'
+pion = 'Pi'
+proton = 'P'
+species = [kaon, pion, proton]
+
```
alexpearce: This makes sense because you need to pass the particle type string to PIDCalib, but I think [this implementation](https://github.com/alexpearce/CharmProduction/pull/54/files#diff-0204676ae12d02f693f2bd014ad0362eR49) could be cleaner.

Add a `pid` argument to the `Particle` constructor:

```patch
diff --git a/analysis/charmproduction/containers.py b/analysis/charmproduction/containers.py
index 8bb8564..59120e7 100644
--- a/analysis/charmproduction/containers.py
+++ b/analysis/charmproduction/containers.py
@@ -65,7 +65,7 @@ class Variable(object):
 
 class Particle(object):
     """Container class representing a particle in a TTree."""
-    def __init__(self, name, title, daughters=None):
+    def __init__(self, name, title, daughters=None, pid=None):
         """Initialise a new Particle object.
 
         Keyword arguments:
@@ -73,10 +73,13 @@ class Particle(object):
         title -- Pretty name to display in TLatex format (default: '')
         daughters -- List of Particle objects representing the decay products
                      of this particle (default: [])
+        pid -- Particle identifier for basic particles, one of config.species
+               (default: None)
         """
         self.name = name
         self.title = title
         self.daughters = daughters or []
+        self.pid = pid
 
     def __getattr__(self, name):
         """Access decay Particle objects with dot notation.
```

Add this information to the `Particle` instances in each mode, e.g. for `DpToKpipi`:

```patch
diff --git a/analysis/charmproduction/ntuples/DpToKpipi.py b/analysis/charmproduction/ntuples/DpToKpipi.py
index dc0681e..08a9ca9 100644
--- a/analysis/charmproduction/ntuples/DpToKpipi.py
+++ b/analysis/charmproduction/ntuples/DpToKpipi.py
@@ -9,9 +9,9 @@ from charmproduction.ntuples.CharmProductionNtuple import CharmProductionNtuple
 class DpToKpipi(CharmProductionNtuple):
     mode = config.DpToKpipi
     head = Particle('Dp', 'D^{#plus}', [
-        Particle('Dp_h1', 'K^{#minus}'),
-        Particle('Dp_h2', '#pi^{#plus}'),
-        Particle('Dp_h3', '#pi^{#plus}')
+        Particle('Dp_h1', 'K^{#minus}', config.kaon),
+        Particle('Dp_h2', '#pi^{#plus}', config.pion),
+        Particle('Dp_h3', '#pi^{#plus}', config.pion)
     ])
     mass_fit_shapes = ('DGS_M', 'FOP_M')
     ipchi2_fit_shapes = ('BGS_IP', 'BGS_IP', 'BGS_IP')
```

Use the `Particle` instance in `pid_cuts` directly:

```patch
pid_cuts = {
    head.Dp_h1: '...',
    # ...
}
```

Then when you loop over `mode.pid_cuts`, you can ask the `Particle` instance directly what type of particle it is. A test should be added/amended to make sure every basic particle in `mode.head` defines this property.

analysis/charmproduction/config.py
alexpearce: OK, sounds good :+1:

analysis/charmproduction/efficiencies/__init__.py
```diff
@@ -7,12 +7,23 @@
...
+    threadpool = Pool(processes=cpu_count())
+    #from charmproduction.utilities import threadpool
+    print 'A'
+    threadpool.apply_async(stripping, [mode, ncheat])
+    print 'B'
+    threadpool.apply_async(selection, args=[mode, bdt_cut, bdt_var])
+    print 'C'
+    threadpool.apply_async(binned_mass_plot, args=[mode, bdt_cut, bdt_var])
+    print 'D'
+    threadpool.close()
```
alexpearce: Please delete the commented out code (git can get it back for us :wink:) and use `log.info` instead, logging something more meaningful (if the logging is necessary).

dominikmuller: That was me saving "wanted to test something" after changing branches. Is already repaired locally.

analysis/charmproduction/pid/GetPIDResultsStatErr.py
```diff
@@ -0,0 +1,455 @@
...
+    if i == j:
+        return ierr*ierr
+    else:
+        jvar = varList[j]
+        # not sure whether I should treat
+        # tracks of same type as 100% correlated,
+        # or tracks of the same type residing in
+        # the same calibration bin
+        #
+        # for the moment, I'll do the former
```
alexpearce: This is addressed in the (hopefully soon to be released) PID calibration note [section 3.2, “Multiple tracks”](https://svnweb.cern.ch/cern/wsvn/lhcbdocs/Users/apearce/PIDCalib_Note/calibration.tex). The conclusion was there's only a correlation if the track types are identical and they fall in the same bin.

analysis/charmproduction/utilities.py
```diff
@@ -7,6 +7,10 @@
 import numpy
 import ROOT
 
+from multiprocessing import Pool, cpu_count
+
+#threadpool = Pool(processes=cpu_count())
```
alexpearce: Delete this too :smile:

analysis/charmproduction/pid/GetPIDResultsStatErr.py
dominikmuller: That part is just copied from pidcalib. But I can change it to what the note says

dominikmuller: And I thought it is done the way you just said, I haven't in detail gone over all the stuff from PIDcalib, thought its implemented the way the note says...

analysis/tests/test_ntuples.py
```diff
@@ -110,6 +114,22 @@ def test_signal_region(self):
             # Signal region defined as (lower bound, upper bound)
             self.assertGreater(sr[1], sr[0])
 
+    def test_daughter_pid(self):
+        """Test if all daugthers have pid set"""
+        for m in MODES:
+            # Need to instantiate the mode to use signal_region
+            inst = m(config.magup, 2011)
```
alexpearce: I don't think you need the instantiation here, because the only member you need to access is `mode.head`, which is a class variable, so you can just use `mode.head.all_daughters()` below.

analysis/tests/test_ntuples.py
```diff
@@ -142,20 +162,22 @@ def test_valid_bdt_variables(self):
 
             # Ignore modes that don't define BDT variables
             try:
-                formulae = map(lambda x: x[0], m.variables)
+                formulae = [v[0] for v in m.variables]
             except AttributeError:
                 continue
 
+            # Validate the various pre and postselections
+            formulae += [m.preselection, m.BDTselection, m.postselection]
```
alexpearce: Should we add the `p_string`, `eta_string`, and `pid_cuts` strings to this list? Because all the variables in those strings should be present.

alexpearce:
Could you please apply this patch to see if it fixes the Sourceforge download problem?

```patch
diff --git a/.travis.yml b/.travis.yml
index c697e82..4013f80 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_install:
 install:
   # Install ROOT using pre-built tarballs
   # https://github.com/alexpearce/travis-ci-root-builds
-  - wget http://downloads.sourceforge.net/project/travis-ci-root-builds/ROOT-${ROOT_VERSION}_Python-${TRAVIS_PYTHON_VERSION}.tar.gz
+  - wget http://tenet.dl.sourceforge.net/project/travis-ci-root-builds/ROOT-${ROOT_VERSION}_Python-${TRAVIS_PYTHON_VERSION}.tar.gz
   - tar xzf ROOT-${ROOT_VERSION}_Python-${TRAVIS_PYTHON_VERSION}.tar.gz
   - mv ROOT-${ROOT_VERSION}_Python-${TRAVIS_PYTHON_VERSION}/root* root
   - source root/bin/thisroot.sh
```

alexpearce:
[This](https://github.com/alexpearce/CharmProduction/pull/54#discussion_r24806278) is the only major outstanding point for me.

dominikmuller:
Yes, I can change that in a second. I also discovered that there is actually a default binning which is recommended for DLLK. I'll put that in instead of the 2010 one. Then it should be done.

alexpearce:
A default ‘recommended’ binning, really? But doesn't the binning depend on the statistics you have available for your signal sample?

dominikmuller:
I don't think so, you compute an efficiency weight per event in your signal sample which obviously does not depend on the statistic in said sample. I am not sure how large the impact of the binning due to the reweighting actually is when averaging over those per event efficiencies. You probably know this better as I do as you worked on PID before.

alexpearce:
I would think that the PID efficiency variation across the binning will vary will the strength of cut, and perhaps a stronger cut will have a larger variation, and should might require a finer binning (to preserve the assumption that the efficiency within a single bin is constant). And of course it should be depend on the statistics of the calibration sample, but generally that is very large.

One useful thing in the PIDCalib ‘default’ binning is that it includes the RICH kaon threshold momenta. The PID efficiency will probably vary quite strongly around those regions, so having a bin boundary there is a good idea.

PR #55, opened 2015-02-03
Tex output module and implementation for efficiencies

dominikmuller:
ToDo:
* Add it for the binned mass plots
* Add it for PID efficiencies once these are merged
* Have to check how this behaves when printing RooFit frames

I created functions for the tex output and modified styles for efficiencies and correlations. Here is a efficiency example: 

out

To have it all work properly, the style has to be set before the histogram is created. It also creates a latex tabular which can be included in documents:

out

 All tex files are automatically processed by pdflatex to create a pdf file. For this to work, `export PATH=/afs/cern.ch/sw/XML/texlive/latest/bin/x86_64-linux:$PATH` is necessary as the old texlive version which is the default one on lxplus cannot handle tikz plots.


alexpearce:
That auto-table is _sweet_. Nicely done.

chrisburr:
Looks good, have you noticed and fixed the `[` in place of `]` issue the the table?

alexpearce:
@chrisburr It's actually [a thing](http://en.wikipedia.org/wiki/ISO_31-11#Sets).

chrisburr:
Ah sorry I forgot about that

dominikmuller:
Whats the `charmproduction/plotting.py` for? Using `grep` I couldn't find anything including it.

alexpearce:
Used in [fitting.py](https://github.com/alexpearce/CharmProduction/blob/bd0349c1be247777b214fd7cb20350b33b177a76/analysis/fitting.py). It's only used for plotting the fits so far, so it would make sense to move it in to this new `plotting/` module.

analysis/charmproduction/plotting/rootTeX.py
```diff
@@ -0,0 +1,64 @@
...
+    retString = retString + texDocument[0]
+    # insert files (wrapped in preview environment)
+    retString = retString + \
+        texInsertion[0] + os.path.abspath(texfile) + texInsertion[1]
+    # end document
+    retString = retString + texDocument[1]
+    return retString
+
+
+def convertTex2Pdf(texfile):
```
alexpearce: for consistency: `convert_tex_to_pdf`

analysis/charmproduction/efficiencies/utilities.py
```diff
@@ -112,7 +115,7 @@ def count_binned(ntuple, mode, pt_name=''):
     return result
 
 
-def draw(before, after, mode, title):
+def calcandprint(before, after, mode, title):
```
alexpearce: for consistency: `calculate_and_print`?

analysis/charmproduction/plotting/canvas_print.py
```diff
@@ -0,0 +1,22 @@
+from charmproduction.plotting.rootTeX import convertTex2Pdf
+import logging as log
+
+from ROOT import TCanvas
+
+
+def DrawCanvas(canvas, name):
```
alexpearce: for consistency: `draw_canvas`, or even something more descriptive like `save_canvas`, as it's actually writing to disk rather than just refreshing it or something.

alexpearce:
After looking over this properly I like it a lot, really nice to have a single plotting module. We should definitely move the current `plotting.py` in here; maybe just move all the utilities methods out in to some `plotting/utilities.py` or whatever, and then major methods in to `plotting/plot_fit.py` or something.

analysis/charmproduction/plotting/rootTeX.py
```diff
@@ -0,0 +1,64 @@
+import os
```
alexpearce: Could lowercase the filename to `roottex` as well :neckbeard:

dominikmuller:
Ok sure, I'll clean up the names.

analysis/charmproduction/efficiencies/selection.py
```diff
@@ -4,7 +4,7 @@
 
 from charmproduction import config
 from charmproduction.efficiencies.utilities import (
-    draw,
+    calculate_and_print as draw,
```
chrisburr: I'd suggest just renaming the instances of `draw` to `calculate_and_print`.

alexpearce:
I've made the migration but the plots look a little worse in my opinion. Maybe it's just because I'm not looking at the LaTeX output though.

what

The shadow around the legend should go and the axis labels for the pull plot are tiny (generally the text should be bigger). Are you basing our style off of the LHCb style guide?

dominikmuller:
The style originated from the LHCb Style once upon the time. But I only put in a simplified version of what I have used before. I'll try to sort out the label size problem. I think I know the cause.

alexpearce:
Good stuff, thanks!

alexpearce:
Better! Still needs a little tweaking, but this is something I can you, or you can look in to it if you like once #63 is merged.

dominikmuller:
I can fiddle around a bit more if you like.

binned

Otherwise I think it is ready to be merged. It also should not break anything already existing. I'll add it to the PID efficiencies once we merged it all.

dominikmuller:
I also left in your version of the style so you can use that as fall back for now.
I am waiting for the integration of the newest ROOT6 version which fixes quite some annoying problems in TTexDump, all of which I already encountered.

dominikmuller:
I just merged the current version of master into this branch and checked that all binning within the selection package is done the same.
I am a bit confused by the large number of changed files if you check the last commit. Shouldn't most of the be identical to master after merging?

chrisburr:
You've actually made a commit which adds all of the changes that have occurred to master since you made this branch hence you have lots of changes. Github has a graph of this [here](https://github.com/alexpearce/CharmProduction/network) or you can look at one of the "merge pull request" commits (e.g. 185e8a0) to see the same thing.

dominikmuller:
So is this bad?

chrisburr:
No I believe its the correct way to be doing things unless Alex wants to jump in and correct me.

The only issue with merging from master like this is you can get double commits if you now merge this into master (like I did with 141d7e662c7de91f7de08f4cebea486cd3ad924d and d6e166d8e282f81e458d30edd399236aede1b9dd) but if you just want to get the latest changes from master its not a problem at all.

You could instead do `git rebase master` to not have a merge commit which makes the history look nicer but sometimes modifying every commit to replay over master nicely just isn't worth the effort.

alexpearce:
I'm not a git pro, but I think it is fine. I generally prefer to rebase, for this reason, but it is really just personal preference. I rebase when either

1. I'm about to close the PR and merge in to master, or
2. Master has been updated with something relevant to the PR. 

For the record, I do

```bash
# Checkout master and pull any new changes from GitHub
$ git checkout master
$ git pull origin master
# Checkout the feature branch and replay its commits on top of the new master
$ git checkout feature-branch
$ git rebase master
# Push the feature to GitHub, forcing any history rewriting
$ git push origin feature-branch -f
```

analysis/charmproduction/efficiencies/binnedmassplot.py
```diff
@@ -0,0 +1,108 @@
...
+        for y in range(n_y_bins):
+            name = '{0}_{1}'.format(y, pt)
+            temp_histos.append(ROOT.TH1F(name, name, 30, mass_low, mass_high))
+        histos.append(temp_histos)
+    # Make a lambda to check the post bdt cuts
+    formula = ROOT.TTreeFormula('post_bdt', mode.postselection, ntuple)
+    formula.GetNdata()
+    # Find the variable names
+    # y_var = mode.rapidity_var.name
+    # pt_var = mode.transverse_momentum_var.name
```
alexpearce: Guess these comments can just be deleted?

analysis/charmproduction/efficiencies/binnedmassplot.py
```diff
@@ -0,0 +1,108 @@
...
+                                       xlow + y * xstep,
+                                       ylow + pt*ystep,
+                                       xlow + (y+1)*xstep,
+                                       ylow + (pt+1)*ystep))
+        pads.append(temp_pads)
+    yields = numpy.zeros((n_pt_bins, n_y_bins))
+    masterhisto.Draw()
+    for pt in range(n_pt_bins):
+        for y in range(n_y_bins):
+            pads[pt][y].SetFillColorAlpha(0, 0.01)
```
alexpearce: Maybe add `pad = pads[pt][y]` and `histo = histos[pt][y]` here to prevent all these list accesses.

analysis/charmproduction/efficiencies/binnedmassplot.py
dominikmuller: Yes, I'll do that once I can log back in

alexpearce:
Looking good, thanks for all the work! :+1:

analysis/charmproduction/efficiencies/__init__.py
```diff
@@ -5,12 +5,14 @@
 from reconstruction import reconstruction
 from stripping import stripping, stripping_by_cut
 from selection import selection
+from binnedmassplot import binned_mass_plot
```
chrisburr: This needs changing to be an [absolute import](https://github.com/alexpearce/CharmProduction/blob/master/analysis/charmproduction/efficiencies/__init__.py) to maintain Python 3 support. You shouldn't need to worry about the others as they should be fixed when you merge/rebase.

dominikmuller:
All done, travis passed

analysis/charmproduction/plotting/roottex.py
```diff
@@ -0,0 +1,64 @@
...
+import subprocess
+import shutil
+import tempfile
+
+
+def make_tex(pkgs, texfile):
+    """Create the content of a tex header file using the packages
+    given in pkgs and inputting the tex-files in files."""
+    texDocHeader = """\documentclass[11pt]{article}
+        \usepackage{blindtext}
```
chrisburr: Do we need `blindtext` here?

analysis/charmproduction/plotting/roottex.py
```diff
@@ -0,0 +1,63 @@
...
+import shutil
+import tempfile
+
+
+def make_tex(pkgs, texfile):
+    """Create the content of a tex header file using the packages
+    given in pkgs and inputting the tex-files in files."""
+    texDocHeader = """\documentclass[11pt]{article}
+        \usepackage[active, tightpage]{preview}
+        \setlength\PreviewBorder{10pt}\n"""
```
chrisburr: You need a double backslash in `\usepackage` else it will be treated as a unicode escape sequence in Python 3.
Also I think this is more readable below but its up to you.
```python
    texDocHeader = ('\\documentclass[11pt]{article}\n'
                    '\\usepackage[active, tightpage]{preview}\n'
                    '\\setlength\\PreviewBorder{10pt}\n')
```

analysis/charmproduction/plotting/__init__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf8 -*-
+"""charmproduction.plotting
+"""
+from .styles import set_efficiency_style
+from .styles import set_correlation_style
+from .styles import set_pretty_style
+from .hist2table import write_latex_table
+from .canvas_print import save_canvas
+from .roottex import convert_tex_to_pdf
+from .plot_fit import plot_fit
```
chrisburr: Python 3 will complain about the lack of `charmproduction.plotting.xxxxx`.

analysis/charmproduction/plotting/roottex.py
alexpearce: I guess you could also use `r'\documentclass[11pt]{article}\n'`, i.e. use a raw string literal.

chrisburr: I thought that had Python 2/3 compatibility issues?

alexpearce: Maybe, I don't have a Python 3 to test, but [according to this](https://docs.python.org/3/reference/lexical_analysis.html) it looks like the `ur` string literal is not supported in 3, but plain `r` is OK.

chrisburr: It does work :+1:

If you ever want to quickly try something in Python 2 or 3 you can use:
http://repl.it/languages/Python
http://repl.it/languages/Python3

PR #56, opened 2015-02-04
Tidy definition of which variables to store locally

chrisburr:
This PR adds the definition of `daughter_branchs` and `res_daughter_branches` in the ntuples classes which are used as prefixes to the variables defined in `CharmProductionNtuples`. If extra variables are required they can be defined as before and will be appended to the automatically generated list.

I've also added a script which can be called and will run the whole framework automatically for the supplied modes but this can be removed if required.


analysis/run_analysis_framework.py
```diff
@@ -0,0 +1,70 @@
...
+parser = argparse.ArgumentParser(description='Run analysis framework')
+parser.add_argument('-m', '--mode', default=None, nargs='+',
+                    help='Mode to run over')
+parser.add_argument('-y', '--year', type=int, required=True,
+                    help='Required: Specifies the year to run over')
+parser.add_argument('-a', '--all', action='store_true', default=False,
+                    help='Run on all available modes (overrides -m)')
+parser.add_argument('-s', '--setup-only', action='store_true', default=False,
+                    help='Only setup directory structure and get data')
+args = parser.parse_args()
```
alexpearce: `argparse` is so sweet.

How about using [`choices`](https://docs.python.org/2/library/argparse.html#choices) for some of the arguments? Something like

```python
parser.add_argument('-m', '--mode', default=None, nargs='+',
                    choices=config.modes, help='Mode to run over')
parser.add_argument('-y', '--year', type=int, required=True,
                    choices=config.years, help='Required: Specifies the year to run over')
```

analysis/run_analysis_framework.py
```diff
@@ -0,0 +1,70 @@
+# This script takes a mode descriptor string and runs all current
+# functionality available in the analysis framework
+
+import argparse
+import logging as log
```
alexpearce: Seeing as progress logging would be pretty helpful here, we could setting the default logging level to `info`. (Optionally adding a quiet `-q` flag or something, but that's probably overkill as most people will want to see the `info` stuff here.)

analysis/run_analysis_framework.py
```diff
@@ -0,0 +1,70 @@
...
+        continue
+    # If we have a 3D mode, copy the weights over from the daughter decay
+    if mode in config.threedim_modes:
+        d_mode = {
+            config.DstToD0pi_D0Tohh: config.D0Tohh,
+            config.DstToD0pi_D0ToKpipipi: config.D0ToKpipipi
+        }
+        weights_path = '{0}/{1}/{2}/{3}'.format(
+            config.output_prefix, d_mode[mode], args.year, config.weights_path
+        )
```
alexpearce: [`os.path.join`](https://docs.python.org/2/library/os.path.html#os.path.join) does this, but I think I normally use `format` too so…

analysis/run_analysis_framework.py
```diff
@@ -0,0 +1,70 @@
...
+    else:
+        log.info('Training the BDT')
+        BDT.train(mode_up)
+        log.info('Producing evaluation plots of the BDT')
+        BDT.evaluate(mode_up)
+    log.info('Producing friend trees of classifier response')
+    BDT.read_all(mode_up)
+    BDT.read_all(mode_down)
+    # log.info('Find optimal BDT cut')
+    bdt_cut = 0.0
```
alexpearce: I guess the optimisation here is to do done later?

alexpearce:
Look at all that red :heart_eyes:

Looks like a much cleaner implementation. I like the script, looks really handy :+1:

analysis/run_analysis_framework.py
chrisburr: Yeah, my thinking was it can be amended as we add features and serve as documentation.

analysis/run_analysis_framework.py
chrisburr: Done.

analysis/run_analysis_framework.py
chrisburr: Done.

analysis/run_analysis_framework.py
chrisburr: That's useful :+1:

chrisburr:
As with #57 I'll merge at the end of the day if there aren't any objections.

alexpearce:
I would like to run this before it's merged, so please bear with my compiler problems… :sob:

chrisburr:
Has there been any progress with this?

alexpearce:
I heard @dominikmuller fixed it but I don't know if the fix has been pushed anywhere?

alexpearce:
OK I've got a fix, just testing the actual script now.

alexpearce:
Right! I get

```
--- <FATAL> DataSetFactory           : Expression TMath::Log10(TMath::Abs(Lc_Loki_BPVVDCHI2)) could not be resolved to a valid formula.
```

and then something crashes (I guess it's the external BDTTools executable that crashes, so the script just carries on regardless).

chrisburr:
@dominikmuller 

I presume this is while downloading data?

chrisburr:
Fixed

alexpearce:
I agree! Now this a few times:

```
Error in <TCling::RegisterModule>: cannot find dictionary module MyDict_rdict.pcm
```

Think it's a problem?

chrisburr:
I also get this but don't know the cause.

alexpearce:
Pushed a sort-of fix for the `TCling` error, think we just need the generated `.pcm` files to be in the same directory as the shared library. Could you double check that BDTTools compiles for you, and the new scripts works as expected?

Then rebase to/merge in master to check that the Travis fix from #61 works on this branch too.

alexpearce:
Or Travis will just work anyway ¯\\\_(ツ)_/¯

dominikmuller:
I get the error as well but tested it and all behaves fine. They are changing the backend for the dictionaries right now so I hope it will get fixed.

chrisburr:
I still get `Error in <TCling::RegisterModule>: cannot find dictionary module MyDict_rdict.pcm` but it otherwise appears to work.

alexpearce:
OK, if it works the same then fine.

In that case, looks like we're good to go :+1:

PR #57, opened 2015-02-04
Calculate correlations

chrisburr:
This uses TMVA to calculate the signal correlation matrix between the provided variables. If no variables are provided it will use all variables in the cheated ntuple that are either floats or doubles except those containing 'TRUE' as TMVA claims that there are an invalid `TTreeFormula`. They work in a `TBrowser` so I have no idea why? We shouldn't need them however so the crude fix can stay unless anyone knows why they cause a problem.


dominikmuller:
Is there an easy way to add a pretty name for the variables so we produce pretty plots?

chrisburr:
Not really, we could have a lookup table added in `TexOutput` and have that as the sole place pretty vars are defined?

analysis/charmproduction/config.py
```diff
@@ -48,6 +48,8 @@
 output_prefix = '../output'
 utilities.ensure_directory_exists(output_prefix)
 weights_path = './weights/'
+temp_dir = '../output/temp'
+utilities.ensure_directory_exists(temp_dir)
```
alexpearce: could use.. an actual [temporary directory](https://docs.python.org/2/library/tempfile.html#tempfile.mkdtemp).

chrisburr: You're full of new python features today

alexpearce: I've been going to sleep listening to an audiobook of the Python API.

chrisburr:
If no one has any objections I'll merge at the end of the day.

analysis/charmproduction/correlations.py
```diff
@@ -0,0 +1,102 @@
...
+    histogram.Draw('textsame')
+
+    t = ROOT.TText(0.53, 0.88, 'Linear correlation coefficients in %')
+    t.SetNDC()
+    t.SetTextSize(0.026)
+    t.AppendPad()
+    c.Update()
+
+    c.Print('{0}/{1}/{2}_correlations.pdf'.format(mode.output_path,
+                                                  mode.polarity, identifer))
```
alexpearce: If we're going for it, `os.path.join` here as well.

alexpearce:
Does this belong better in `charmproduction.plotting`, perhaps?

chrisburr:
I agree but in the current state of this that doesn't exist and clashes with `plotting.py`. I suggest we move it when `TexOutput` is ready.

alexpearce:
:+1: sounds good! that's all from me.

PR #58, opened 2015-02-05
Add sensible mass, IP χ², and δm binnings.

alexpearce:
Taken from the 2010 note. Pretty standard stuff I think. We will of course add the transverse momentum and rapidity binnings in the same place, but I've left them out for now as we'll decide on that later.


chrisburr:
Looks good, the [efficiencies](https://github.com/alexpearce/CharmProduction/blob/master/analysis/charmproduction/efficiencies/utilities.py#L42) should also be updated.

alexpearce:
For sure, shall I do that here or should one of us open another PR for it?

chrisburr:
I think it fits here but if you'd rather it was separate go ahead and merge and I'll make a new one.

alexpearce:
OK, I had a quick look at the usage in `efficiencies` but there's a little more to replace. Would you mind taking a look if I merge this?

chrisburr:
Sure

alexpearce:
Thanks :+1:

PR #60, opened 2015-02-05
Add 2010 pt and y binning

chrisburr:
Also fixes a bug I introduced changing to `os.path.join` which required the year to be passed as a string.


alexpearce:
Ah very good, I already had a ‘patch’ locally to fix that one :wink:

Could you rebase with `master` and use [`uniform_binning`](https://github.com/alexpearce/CharmProduction/blob/c8967d353fad96047999817551393548fa740da3/analysis/charmproduction/utilities.py#L76) instead of `range`?

alexpearce:
All looks good to me.

chrisburr:
All done I believe

analysis/charmproduction/efficiencies/utilities.py
```diff
@@ -40,17 +40,15 @@ def get_real(mode, bdt_var=None):
...
 
 
 def get_y_binning(mode):
-    y_min = 2.0
-    y_max = 4.5
-    n_y_bins = 5
+    y_min, y_max = mode.rapidity_var.binning_range()
+    n_y_bins = mode.rapidity_var.binning.numBins()
     y_bin_size = (y_max-y_min)/n_y_bins
     return y_min, y_max, n_y_bins, y_bin_size
```
alexpearce: (Sorry, nearly there.) Is it/will it be a problem if the bins aren't equal width?

chrisburr: (Don't be) Currently yes, because of how it determines which bin an event belongs in (here)[https://github.com/alexpearce/CharmProduction/blob/master/analysis/charmproduction/efficiencies/stripping.py#L50].

alexpearce: If you use to the `RooBinning` object (`whatever_var.binning`), then you could use [`binNumber`](https://root.cern.ch/root/html600/RooBinning#RooBinning:binNumber)?

analysis/charmproduction/efficiencies/selection.py
```diff
@@ -26,8 +23,8 @@ def selection(mode, bdt_cut, bdt_var=config.default_bdt_var):
     # Loop over all candidate events
     for e in ntuple:
         # Find the index in y/pt space of the event
-        y = int((ntuple.val(y_var)-y_min)/y_bin_size)
-        pt = int((ntuple.val(pt_var)-pt_min)/pt_bin_size)
+        y = mode.rapidity_var.binning.binNumber(ntuple.val(y_var))-1
+        pt = mode.pT_var.binning.binNumber(ntuple.val(pt_var))-1
```
alexpearce: I thought I commented on this but must have refreshed the page.

Member lookup shouldn't take _too_ long, but inside a loop we should keep things snappy. So could do

```python
pT_binning = mode.pT_var.binning
for e in ntuple:
    pt = pT_binning.binNumber(ntuple.val(pt_var)) - 1
```

or even one step further.

```python
pT_find_bin = mode.pT_var.binning.binNumber
for e in ntuple:
    pt = pT_find_bin(ntuple.val(pt_var)) - 1
```

alexpearce:
I'll look in to binNumber rounding the bin number in to the lowest and highest indices tomorrow. Just to be sure, what behaviour do you want for x values outside the bin range, bin numbers outside {0, N} (where N is the number of bins), an exception, or something else?

chrisburr:
Locally I've added this to `Variables` which then removes the need for the `try` statement and makes the using it much more compact.
```python
    def bin_number(self, ntuple):
        val = ntuple.val(self.name)
        if self.binning.lowBound() <= val <= self.binning.highBound():
            return self.binning.binNumber(val)
        else:
            return None
```
What are your thoughts?

dominikmuller:
I wouldn't mind keeping the things outside the range in under and overflow bins 0 and N+1. However, this opens up for a wide range of indices screw ups by us so probably just drop them.
Can we still easily get a list of bin boundaries? I need to build bin cutstrings to cut the tree for PID efficiency calculation.

alexpearce:
I agree with @dominikmuller that returning any integer for under/overflow bins could lead to a bug down the line; I initially thought of returning -1 for x values outside the binning, but then if you're using that as a list index you'll incorrectly get the last element. Returning None might be the best option here.

@chrisburr For the implementation I would prefer in `Variable.bin_number` was passed `val` directly, rather than an `Ntuple`. Seems a bit cleaner to me (and maybe it's cheaper passing around numbers rather than objects). Also shouldn't it be `< highBound` rather than `<= highBound`?

We can get the list of bin boundaries using [`RooBinning.array`](https://root.cern.ch/root/html600/RooBinning#RooBinning:array), but it is a little ugly as you have to ‘convert’ a `double*` to a Python `list`. See [this test](https://github.com/alexpearce/CharmProduction/blob/master/analysis/tests/test_variable.py#L58) for an example of that. We could add `Variable.bin_boundaries` to automate this.

chrisburr:
The motivation for passing ntuple instead of the value is then we don't need to get the find the name ([like here](https://github.com/alexpearce/CharmProduction/blob/correctly-bin-efficiencies/analysis/charmproduction/efficiencies/selection.py#L21)) and we don't risk switching the names accidentally.

I'll have a look at profiling it to see if it matters.

dominikmuller:
@alexpearce In ROOT histograms iBIn = 0 is the underflow and iBin = N+1 is the overflow bin. The first "normal" bin is index 1. But it is still inviting errors on our end which are easily made but hard to find.

alexpearce:
@chrisburr OK. I still prefer passing the value directly to avoid coupling the two objects. The only other thing I can think of to tidy things up would be

```python
# Find the variable names
y_ref = ntuple.val(mode.rapidity_var.name, reference=True)
pT_ref = ntuple.val(mode.pT_var.name, reference=True)
# Cache the method that finds the bin number (whatever one we use)
y_find_bin = mode.rapidity_var.binning.binNumber
pT_find_bin = mode.pT_var.binning.binNumber
# Loop over all candidate events
for e in ntuple:
    # Find the index in y/pt space of the event
    y = y_find_bin(y_ref[0]) - 1
    pt = pT_find_bin(pT_ref[0]) - 1
```

Then you don't have to keep looking up the value by name. (Under the hood this effectively returns a pointer to the value of the branch for the current TTree entry, which you ‘dereference’ with `ref[0]`.) So, it's a little ugly cos you're sort of going back to the C++ way of doing things.

I'll leave it up to you to decide what's best, it is only a minor point, really.

chrisburr:
`reference=True` [doesn't do anything](https://github.com/alexpearce/CharmProduction/blob/master/analysis/charmproduction/Ntuple.py#L100) so I'll remove the arugument from `Ntuple.val` and use `ref` instead.

alexpearce:
Haha brilliant. Good spot! Hooray for comprehensive testing...

chrisburr:
Removing the name lookup from the loop saves around 1 second per minute, switching to ref halves the time it takes. I'm sold.

alexpearce:
Halves the time? Holy crap. Yeah nice one.

alexpearce:
We should get this merged soon so I can stop fixing the `year` bug in all my branches :stuck_out_tongue:

chrisburr:
You can pass the year as a string when initialising and everything works :smile:

alexpearce:
Yeah I've applied that patch to each branch of master, so once this is merged I can rebase everything.

chrisburr:
Unless I've missed something this is done

alexpearce:
Doe [this comment](https://github.com/alexpearce/CharmProduction/pull/60#issuecomment-73127182) still need addressing?

chrisburr:
I think [this](https://github.com/alexpearce/CharmProduction/blob/3d92278fd3b5075539035a009a7a0e5a1c4b2f7e/analysis/charmproduction/containers.py#L47) deals with it by just returning none.

I'll fix the `<=` momentarily.

alexpearce:
Oops sorry, I had read that but just forgot about it.

Otherwise all god :+1: Thanks for all the work!

chrisburr:
I'll merge shortly if someone (looking at you travis) doesn't complain.

alexpearce:
:rocket:

PR #62, opened 2015-02-06
Fix Cling .pcm warning

alexpearce:
By symlinking `MyDict_rdict.pcm` to inside `bin/` and `lib/`, neither using the executable nor the shared library result in the Cling error.


chrisburr:
:+1:

PR #63, opened 2015-02-06
Fitting script

alexpearce:
Tweaks to the fitting script to make it easier for people to run and to use the same output directory as everything else. I'll change the [`get_ntuple`](https://github.com/alexpearce/CharmProduction/blob/692a8c1cedb4e644a708164fc9ef5d3a218b9514/analysis/fitting.py#L41) method to use the processed data in `../output/`, rather than the curernt hardcoded paths for 2010 data.

```shell
$ ./fitting.py -m DsTophipi -p MagDown -y 2011
```


alexpearce:
This should be good to go. You will need to apply [this patch](https://github.com/alexpearce/CharmProduction/commit/f39419648ab6840e60abb693a3807bc86f4546c1#diff-2388caeed931ee8a5f512eedbd546e03R91) if you want to test yourself (encouraged!).

chrisburr:
I've forgotten my charger so I email or GitHub me if you want my attention.

Seems to work, is all [this](https://gist.githubusercontent.com/chrisburr/e458da2961027723dce3/raw/log.txt) output normal?

chrisburr:
Also am I correct in thinking this isn't actually applying any selection to the ntuples?

alexpearce:
It is not normal output (thanks for testing)! It's because the negative IPCHI2 values aren't being removed, so the logarithm returns NaN. I'll push a fix.

That's right, there's no selection applied. I'm trying to think of the nicest way to apply it.

alexpearce:
@chrisburr Care to try again?

chrisburr:
From what I can tell the log looks normal exit for printing this periodically while plotting:
`WARNING: Could not find parameter sigmaR_ip_bkg, skipping`
`WARNING: Could not find parameter sigmaR_ip_sec, skipping`
If that isn't an issue though it's looking really nice and merge whenever.

alexpearce:
Does [`b9e31a`](https://github.com/alexpearce/CharmProduction/commit/b9e31a6bca955b66dcd482f59491aa9f8150ae17) answer your selection points?

And yes, the `Could not find parameter...` stuff is harmless.

alexpearce:
The (oddly new) Travis errors are unrelated to this PR. I'll fix them in another one.

alexpearce:
@chrisburr Selection looking OK here? Think that will be everything.

chrisburr:
Sorry I thought I had already answered this, all looks good to me.

PR #64, opened 2015-02-06
Central dictionary for pretty variable names and units

dominikmuller:
Based on Chris' suggestion I did a first implementation for a central pretty names dictionary. It uses a dict mapping the actual names to the pretty name and unit:

```
'ThirdLead*PT': ('3$^#text{rd}$ Lead. $p_{T}$', 'MeV')
```

Both the unit and the variable should be formatted the way you would usually do it in LaTeX without the need to place decorating stuff  to trick ROOT tex output. The only exception is the need to use `#` instead of `\` for LaTeX commands.
There are for different functions to access the pretty names depending on what's needed. Each provided actual name is compared to all keys in the dict by determining which keys are substrings of the provided actual name. Of those, the longest matching substring is chosen to determine the pretty name.
Keys including `*` are split and all resulting strings have to be a substring of the actual name.

See below for the changes I did to the BDT wrapper functions and the DpToKpipi mode. If both of you are happy I am going to modify all other modes and add their variables to the dict.


analysis/charmproduction/prettyvar.py
```diff
@@ -0,0 +1,75 @@
...
+    'ETA': ('$#eta$', ''),
+    'PT': ('$p_T$', 'MeV'),
+    'Y': ('$y$', ''),
+    'E': ('$E$', ''),
+    'TRUE_P': ('$p$', 'MeV'),
+    'Lead*PT': ('Lead. $p_{T}$', 'MeV'),
+    'SecLead*PT': ('2$^#text{nd}$ Lead. $p_{T}$', 'MeV'),
+    'ThirdLead*PT': ('3$^#text{rd}$ Lead. $p_{T}$', 'MeV'),
+    'DIRA': ('DIRA', ''),
+    'Loki': ('$#cos(#theta_{#pi})$', ''),
```
chrisburr: Shouldn't this be 'Loki_LV02'

dominikmuller: Yes it should. My mistake. Thanks!.  Forgot to add that the actual content of the dict is no final.

chrisburr:
Can something like `D0_PY` be confused with `D0_Y`? And the `CHI2*NDOF will pass for two different variables I think ([see here](https://github.com/alexpearce/CharmProduction/blob/master/analysis/charmproduction/ntuples/CharmProductionNtuple.py#L56))

Otherwise looks good so feel free to finish it.

dominikmuller:
Should be fine as the longest matching string would be PY and not Y.

PR #66, opened 2015-02-07
Add paths and fix D0 variables for the Dst modes

chrisburr:
This is still missing a path somewhere but its mostly ready.


alexpearce:
Thanks! I would like to add a test that checks that all the variables required by the derived classes, like `D0Tohh`, is present somewhere, either in `CharmProductionNtuple.mother_vars` or similar. Just working on it now.

chrisburr:
I think this now work so merge whenever you're happy.

alexpearce:
Should the `Dst` modes have a `variable` member defining the BDT inputs?

chrisburr:
Yes it should, I'll fix it in a moment

chrisburr:
Everything should be fixed. If you have any D0 data downloaded it needs redownloading after b74443c as the MagUp MC was actually MagDown.

alexpearce:
I've added the tests in #68 so I think this can be merged.

PR #68, opened 2015-02-09
Add tests for CharmProductionNtuple classes

alexpearce:
This was motivated by #66.

These tests should make it more obvious when a child class of `CharmProductionNtuple` is missing properties required by the framework. For example, it [checks](https://github.com/alexpearce/CharmProduction/commit/86c6818b96f71daf86dd28651d64ba0843b21ceb#diff-65e133ecf76ea190d347c86c39eaf0cdR119) that all BDT training variables are copied locally when `mode.setup_data` is called, and that each mode defines the [correct number of valid PDFs](https://github.com/alexpearce/CharmProduction/commit/86c6818b96f71daf86dd28651d64ba0843b21ceb#diff-65e133ecf76ea190d347c86c39eaf0cdR78).


alexpearce:
So here's the nice thing about tests :smile: 

[This test](https://github.com/alexpearce/CharmProduction/blob/86c6818b96f71daf86dd28651d64ba0843b21ceb/analysis/tests/test_ntuples.py#L60) spotted that `DstToD0pi_D0ToKpipipi` doesn't define `daughter_branches` (it sort of does, but there's just [a typo](https://github.com/alexpearce/CharmProduction/blob/86c6818b96f71daf86dd28651d64ba0843b21ceb/analysis/charmproduction/ntuples/DstToD0pi_D0ToKpipipi.py#L10)).

[This test](https://github.com/alexpearce/CharmProduction/blob/86c6818b96f71daf86dd28651d64ba0843b21ceb/analysis/tests/test_ntuples.py#L119) spotted that `D0_Loki_DOCAMAX` is [used in the BDT training](https://github.com/alexpearce/CharmProduction/blob/86c6818b96f71daf86dd28651d64ba0843b21ceb/analysis/charmproduction/ntuples/D0ToKpipipi.py#L22) but is [not copied locally](https://github.com/alexpearce/CharmProduction/blob/86c6818b96f71daf86dd28651d64ba0843b21ceb/analysis/charmproduction/ntuples/CharmProductionNtuple.py#L56-L59) during `setup_data`. This has actually been fixed in #66. I'd like to get that merged before this, to check the relevant test passes.

chrisburr:
I'll get around to writing some at some point :smile:

Go ahead and rebase when you're ready.

alexpearce:
I'm a little confused by some of the new member variables, for example in `DstToD0pi_D0ToKpipipi`

```python
mother_branch = 'Dst'
mother_latex = 'D^{*+}'
extra_mothers = ['D0']
extra_mothers_latex = ['D^{0}']
daughter_branches = ['pi']
res_daughter_branches = ['D0_h1', 'D0_h2', 'D0_h3', 'D0_h4']
```

Will `res_daughter_branches` always be daughters of `extra_mothers`? And `daughter_branches` is only "simple particle" daughters of `mother_branch`?

I'm wondering if some some of dictionary structure might be better, because once we get `DsstToDsgamma_DsTophipi` we'll end up with two decays (the D_s*+ and the phi).

chrisburr:
Ah I hadn't realised we had a decay that was three levels deep. I originally set the name with `res_daughter_branches` standing for the resonance daughters, though it might be nicer to only use `daughter_branches` and specify `Dst_pi` rather than adding the mother name automatically.

PR #71, opened 2015-02-10
Fix tex output

chrisburr:
This PR fixes the segfault I saw when initialising the `TH2F` object in `charmproduction.plotting.utilities`.


alexpearce:
I don't understand why this would segfault as this is very allowed:

```python
import ROOT
b = ROOT.RooBinning(0, 1)
b.addBoundary(0.3)
b.addBoundary(0.5)
b.addBoundary(0.8)
b.addBoundary(0.9)
h = ROOT.TH1F('h', 'h', b.numBins(), b.array())
```

Maybe you don't need to call `RooBinning.array.SetSize` if you're passing the array to a ROOT method? I only call `SetSize` when I want to convert the `double*` array to a Python `list`.

chrisburr:
I have no idea why it happens but the snipped you've written also segfaults when initialising the `TH1F`.

alexpearce:
Interesting, it doesn't for me. Python 3 thing? Does it segfault on lxplus?

chrisburr:
Its happy on lxplus so it must be.

dominikmuller:
So should we merge this? I only tested it on lxplus so didn't have any problems.

chrisburr:
That's fine, go ahead and merge both branches when you're ready.

alexpearce:
I would have actually preferred more work on this as it's not really obvious why the `array` module needs to be used, looks like a Python 3 (+ ROOT) bug rather than a necessary feature. Plus having a `bins` _and_ a `binning` member variable is super confusing.

If we would rather use this fix than something else, I'd at least suggest a more explicit name for the member variable, like `binning_as_array` or `array_binning` (and a comment explaining why it's needed).

alexpearce:
Also this could cause the two binnings (!) to get out of sync, if you assign something to `bins` but not to `binning`.

PR #72, opened 2015-02-10
Move containers in to their own module.

alexpearce:
They were becoming numerous, and will become moreso with #70, so have one file
per container.

The tests have also been moved in to a [`tests/containers`](https://github.com/alexpearce/CharmProduction/tree/5d2df4ba47a594d818399256c3510abc4132089f/analysis/tests/containers) folder.


chrisburr:
:+1:

Would it make sense to move `Ntuple.py` to this module?

alexpearce:
Yes!

alexpearce:
@dominikmuller I've been writing the `Particle` titles in `TLatex`, like `#pi_{#text{Soft}}^{#plus}`, should I be using regular LaTeX instead, for the .tex output stuff?

chrisburr:
I think they need to be regular latex with `\` replaced with `#` as [here](https://github.com/alexpearce/CharmProduction/blob/08ffe7c7fc2a105496ab56ecf9ad2ca71c3d3354/analysis/charmproduction/prettyvar.py#L57).

chrisburr:
Is there anything preventing this being merged?

alexpearce:
Nope!

Issue #74, opened 2015-02-10
Bug in DstToD0pi_D0ToKpipipi BDT selection?


alexpearce:
Is this a bug?

```python
BDTselection = ('Dst_ENDVERTEX_CHI2 < 16 && D0_pi_TRACK_CHI2NDOF < 5')
```

Shouldn't it be this?

```patch
diff --git a/analysis/charmproduction/ntuples/DstToD0pi_D0ToKpipipi.py b/analysis/charmproduction/ntuples/DstToD0pi_D0ToKpipipi.py
index 89fb154..12e640b 100644
--- a/analysis/charmproduction/ntuples/DstToD0pi_D0ToKpipipi.py
+++ b/analysis/charmproduction/ntuples/DstToD0pi_D0ToKpipipi.py
@@ -20,7 +20,7 @@ class DstToD0pi_D0ToKpipipi(CharmProductionNtuple):
     depth = D0ToKpipipi.depth
     variables = D0ToKpipipi.variables

-    BDTselection = ('Dst_ENDVERTEX_CHI2 < 16 && D0_pi_TRACK_CHI2NDOF < 5')
+    BDTselection = ('Dst_ENDVERTEX_CHI2 < 16 && Dst_pi_TRACK_CHI2NDOF < 5')

     postselection = ('D0_h1_ProbNNk > 0.1 && D0_h2_ProbNNk < 0.1 && '
                      'D0_h3_ProbNNk < 0.1 && D0_h4_ProbNNk < 0.1')
```

And the [same for `DstToD0pi_D0Tohh`](https://github.com/alexpearce/CharmProduction/blob/0e11821aca73bf81d4eeeed4d8c8b7e3c561afeb/analysis/charmproduction/ntuples/DstToD0pi_D0Tohh.py#L33).

chrisburr:
Yes it is. It must not have been used as it shouldn't even compile that formula.

alexpearce:
I came across it running the analysis script with `--all`.

alexpearce:
I'll make the changes and open a PR.

alexpearce:
Closed by merging #75.

PR #75, opened 2015-02-10
DstToD0pi typos

alexpearce:
Fixed the cut string typos discussed in #74 and added the pre and postselection cut strings to the [branch name validity tests](https://github.com/alexpearce/CharmProduction/blob/c567f206f7f183e3a0cccae13730140538d759c5/analysis/tests/test_ntuples.py#L104).


alexpearce:
@chrisburr Look about right?

chrisburr:
Go! :+1:

PR #77, opened 2015-02-10
Fix the lists of variables which are downloaded by setup_data()

chrisburr:
Well I now understand how maps work :smile:


alexpearce:
Yep! Sorry I didn't think of this, I had stumbled across the ‘bug’ using `map` in the tests, but didn't check its usage elsewhere.

Merge!

PR #78, opened 2015-02-10
Rename D0Tohh to D0ToKpi

alexpearce:
This must not be merged until new ntuples have been produced with the “D0ToKpi” naming. With these changes, this will be done the next time the options files are used.

I haven't yet altered the ntuple options to only select K-pi+ D0 decays from the stripping, though. Should I do that in this PR?


chrisburr:
You can or I can do it when modifying everything for Stripping 21?

analysis/charmproduction/ntuples/DpToKKpi.py
```diff
@@ -92,10 +92,10 @@ def set_data_vars(self):
         value = (cb_real + '11_MagUp/{0}/DVntuple.root', 1, 16, 310)
         self.add_to_data_paths(2011, 'Real', 'MagUp', value)
 
-        value = (cb_mc + '15_MagDown_D0Tohh/{0}/DVntuple.root', 1, 2, 10)
```
chrisburr: Should this be here?

alexpearce: I was wondering about this line, it was here when I edited it :smile: But I don't know if it's incorrect, I didn't check that directory. There are other non-D0 files in this commit with the same changes.

chrisburr:
I take that back, seeming as you've modified the options you might as well modify the decay descriptor so it doesn't get missed.

analysis/charmproduction/ntuples/DpToKKpi.py
chrisburr: It is incorrect, I'll look at fixing it when I've ran DaVinci again.

alexpearce:
Yep good point.

alexpearce:
I have checked the results of [`c82ac50`](https://github.com/alexpearce/CharmProduction/commit/c82ac5078ad976c84190f6fabcde8ad3d205c896) by comparing the number of entries in the ntuples before and after the changes. I selected K- pi+ and K+ pi- entries from the original options, using `D0Tohh`, with the cut string

```
(abs(D0_h1_ID) == 321 && abs(D0_h2_ID) == 211) ||
(abs(D0_h1_ID) == 211 && abs(D0_h2_ID) == 321)
```

For all ntuples in both collision and MC the number of entries is identical to that of the new `D0ToKpi` options, and the distributions of the selected entries are also the same.

chrisburr:
I've made a symlink for the MC data paths so both D0Tohh and D0ToKpi are valid, so I guess this can be merged?

alexpearce:
Isn't it still a problem that the TTree is called D0Tohh?

chrisburr:
It is I forgot about that :disappointed:

PR #80, opened 2015-02-12
RooFit Plots + Tex output

dominikmuller:
pretty

Still some work left, have to organise the actual saving of all the plots somewhere. Also haven't checked any other but the integrated invariant mass plot yet.


analysis/charmproduction/plotting/utilities.py
```diff
@@ -110,9 +112,20 @@ def add_fit_parameters(canvas, params_names, workspace, show_consts=True):
...
+    box = ROOT.TPaveStats(
+        0.65,
+        topbdr -
+        0.05 *
+        nparam,
+        rightbdr -
+        0.01,
+        topbdr -
+        0.01,
+        'NDC')
```
alexpearce: Could you change the layout of this slightly?

```python
box = ROOT.TPaveStats(
    0.65, topbdr - 0.05*nparam, rightbdr - 0.01, topbdr - 0.01, 'NDC'
)
```

analysis/charmproduction/plotting/utilities.py
```diff
@@ -120,14 +133,19 @@ def add_fit_parameters(canvas, params_names, workspace, show_consts=True):
...
+            const = '#text{ (C)}'
         else:
             const = ''
-        latex.DrawLatexNDC(
-            0.65,
-            0.85 - (0.05*idx),
-            '{0} = {1:.2f} #pm {2:.2f} {3}'.format(name, val, err, const)
-        )
+        label = '{0}{3}= {1:.2f} #pm {2:.2f}'.format(name, val, err, const)
+        print label
```
alexpearce: `log.info`?

alexpearce:
This is looking really nice :boom: :art: :boom:

Only a couple of things from me.

1. Data points should be larger, maybe a [`MarkerSize`](https://github.com/alexpearce/CharmProduction/pull/80/files#diff-e1c7b21b5d3836faf8448f1e75cc2de2R161) or 1 or even 2?
2. I prefer a non-square [canvas size](https://github.com/alexpearce/CharmProduction/pull/80/files#diff-0cd818b79f1b30a6bafdee0b41d8f543L86). As it's a two-part plot, I think it looks better to be taller than a square its not too squashed vertically.

For the saving, what would be ideal for me is to have these styles applied, saved as usual to the `fits.root` file, and then be exported to .tex when that's needed. Is that possible? If they're saved to .root now, then .tex, does that screw up the latex, rather than saving directly to .tex?

analysis/charmproduction/plotting/utilities.py
dominikmuller: Sorry sure, that's what my vim did automatically

analysis/charmproduction/plotting/utilities.py
dominikmuller: That was just me trying to figure out why the box didn't work. Will be removed

dominikmuller:
Yes sure, I will not break anything off the existing stuff. Due to the way tex labels work, stuff will look weird if you don't save as tex and pdf latex it. I thought there was style guide saying you should use 800x600 or 800x800, but maybe that was atlas. 
I am working on setting up a thread pool to which the pdf latex commands are dispatched so you should get the pdfs right away without performance impacts.

alexpearce:
Off of the top of my head, I don't think there's a recommendation on image size in the LHCb style guide. Pull plots often aren't shown in published stuff though, so we can use what we like internally.

Yeah I know that's the tradeoff of using the latex stuff in the labels, it looks funny in a TBrowser. That's OK, I can live with that :smile:

analysis/fitting.py
```diff
@@ -254,12 +258,26 @@ def fit(mode, polarity, year):
...
+                tot_pdf.pdf,
+                'pdf_m_dm_ip_sec'),
+            FitPDF(
+                'Secondary random soft pion',
+                tot_pdf.pdf,
+                'pdf_m_dm_ip_sec_d0'),
+            FitPDF(
+                'Combinatorial background',
+                tot_pdf.pdf,
+                'pdf_m_dm_ip_bkg')]
```
chrisburr: Is there a reason for these lines changing or is it just vim?

alexpearce:
Thanks! And now for something controvesial…

How do people feel about centered axis labels? I quite like them :grin:

dominikmuller:
Oh boy... I would have to see the plots first before having an opinion. But I doubt it :P

alexpearce:
Haha OK, it is a very personal preference. It seems much more common outside LHC stuff, it's the default in Python, but it's probably best with stick with right-aligned titles anyway to prevent rattling too many cages…

alexpearce:
Example of centred labels, for... information... 

c_mass_integrated

dominikmuller:
Also looks nice. I am indifferent, we can use whatever you prefer.

chrisburr:
I prefer it also, especially on the pull plot axis.

analysis/charmproduction/plotting/styles.py
```diff
@@ -25,7 +25,9 @@ def set_pretty_style():
 
     # Font scaling.
     prettyStyle.SetLabelFont(43, "xyz")
-    prettyStyle.SetLabelSize(20, "xyz")
+    prettyStyle.SetLabelSize(25, "xyz")
+    prettyStyle.SetTitleFont(43, "xyz")
+    prettyStyle.SetTitleSize(30, "xyz")
```
alexpearce: This might be my fault because I copied a style from an older project using double quotes `"`, but this file should be using single quotes `'` for consistency.

analysis/charmproduction/plotting/utilities.py
```diff
@@ -57,9 +57,11 @@ def add_pull_plot(canvas, plot_var, frame):
     # Prepare the canvas for the pull plot
     canvas.Divide(1, 2, 0.01, 0.05)
     canvas.cd(1)
-    ROOT.gPad.SetPad(0, 0.25, 1, 1)
+    # ROOT.gPad.SetBottomMargin(0.1)
```
alexpearce: Delete line.

alexpearce:
I'd like to get this good stuff merged! The only stuff left to address are the [two](https://github.com/alexpearce/CharmProduction/pull/80#discussion-diff-24807733) [comments](https://github.com/alexpearce/CharmProduction/pull/80#discussion_r24807744) right?

alexpearce:
Sweet! Looks good to me :+1: 

Thanks for all the work :star:

PR #82, opened 2015-02-13
Uncertainties on efficiencies using Agresti-Coull intervals

dominikmuller:
I changed the computation of the uncertainties on the efficiencies using to the TEfficiency implementation of the Agresti-Coull intervals. Those looked pretty simple to me and fix the issue of significantly underestimating the actual uncertainty for very high and very low efficiencies in case of small statistics.

The plots take the maximum difference of the upper and lower bound w.r.t. to the nominal value as the printed error.


chrisburr:
That's a much nicer way of doing it.

alexpearce:
I agree that it's nicer to use something existing than to make our own.

One hesitation I have is due to my lack of understanding: why "Agresti-Coull" intervals? I've not heard of this, I've only ever computed the "normal" binomial uncertainty on efficiencies. Can we motivate this choice?

Secondly, if we're using `TEfficiency` should we go all-in, using the [`Fill`](https://root.cern.ch/root/html/TEfficiency.html#TEfficiency:Fill) and [`Draw`](https://root.cern.ch/root/html/TEfficiency.html#TEfficiency:Draw) methods to have `TEfficiency` create the histogram for us?

dominikmuller:
First, I do like the name of this method. We could also use Feldman-Cousin, but this seemed sounds like a specialised confidence level thing for binomial statistics.

http://www.jstor.org/stable/2685469

I looked over the paper and will read it over the weekend, can tell you more on monday. Lorenzo told me that he used this method before and that it was utilized in Barbar publications. I was thinking about doing that. Just have to check how well this can be combined with the normal plotting functions as TEfficiency does not inherit from TH1, well see. would also be great for storing the efficiencies. I'll test it later or over the weekend.

dominikmuller:
The tables can get asymmetric errors though

chrisburr:
Do you know if it is possible to put asymmetric errors on the plots?

dominikmuller:
TGraphAsymErrors or something like that can do that, don't know about histograms and I highly doubt it in case of the TEXTe plotstyle

alexpearce:
Thanks for the paper link, I'll have a read too.

alexpearce:
I think this can be merged :+1:

PR #84, opened 2015-02-20
Fitting fine tuning

alexpearce:
This PR contains mostly minor changes to the fitting framework, allowing it to fit all data samples with decent performance and making the fitting across the modes more consistent.

There are a couple of larger changes conceptually.

### sWeights

The background data sample used to fit the background IP χ² shape is no longer the sidebands from the mass distribution in collision data, but the collision data weighted by background sWeights from a 1D invariant mass fit.

This is nice because you can remove any tails from e.g. Crystal Balls on on statistical basis from the background sample, and because you can effectively use the full background “statistics”, unlike using sidebands where you don't use the data in the signal window.

### IP χ² signal shapes

After investigating the possible IP χ² shapes, it looks mostly obvious to me that the best choice for the signal shape is the asymmetric bifurcated Gaussian with exponential tails. This is the same `AsymGaussExp` shape as used in 2010. I have also allowed the mean of this PDF to float across pT-y bins as the signal MC clearly varies in this parameter across pT and y.

Tentatively, I've also allowed the background shapes to be asymmetrical. The 2010 note constrainted both the background and secondary shapes to be symmetrical, but I think I can see a good case for dropping the former.

### D⁰ → K⁻π⁺π⁻π⁺

There is still the problem that @chrisburr found of the high IP χ² tail in the `D0ToKpipipi` collision signal distribution. I haven't come up with an explanation, and it worries me that the distributions looks like the sum of at least two components, which is unlike any other mode. My only suspicion is some physics background causing it, like a secondary three body decay being combined with a random kaon/pion, peaking in the signal region. Needs more investigating.

I'll present all this next week, and if everyone's OK with the changes we can merge.


chrisburr:
Should we have an argument in `fitting.py` to enable sweights?

Also should [this](https://github.com/alexpearce/CharmProduction/pull/84/files#diff-71c9d35b91578eceb6414c6421bd0329R66) and [this](https://github.com/alexpearce/CharmProduction/pull/84/files#diff-ef39e133aa7dee62ee1be73592d9597eR76) be split across three lines?

alexpearce:
Adding an sWeight argument is a good idea.

I formatted like that to match the style of the `unfix_conditions` list just above it, so you can quickly scan all the variables that are being split. As there's only one in this case it might be redundant.

alexpearce:
I've added [a Makefile](https://github.com/alexpearce/CharmProduction/blob/95bfa54a98df039101eb9a2a44d98e70e01cb160/analysis/charmproduction/fitting/shape_classes/Makefile) to [allow compilation](https://github.com/alexpearce/CharmProduction/commit/95bfa54a98df039101eb9a2a44d98e70e01cb160) of the [custom shape classes](https://github.com/alexpearce/CharmProduction/tree/95bfa54a98df039101eb9a2a44d98e70e01cb160/analysis/charmproduction/fitting/shape_classes). The benefit of doing this, as opposed to just dynamically loading the `.cxx` files with `gROOT.ProcessLine('#include "MyClass.cxx"')`, is that the classes can be persisted in ROOT files, allowing later inspection (though you need to [load the library object](https://github.com/alexpearce/CharmProduction/blob/95bfa54a98df039101eb9a2a44d98e70e01cb160/analysis/charmproduction/fitting/ipchi2.py#L325) before accessing saved objects).

@chrisburr and @dominikmuller, could you please just check that you can run `make` without any problems on your machines?

dominikmuller:
Good job. I'll check it tomorrow.

chrisburr:
Works for me :+1:

dominikmuller:
Works for me as well. I just noticed that you only implemented `evaluate()` and the analytical integral. I had an experience once in `RooFit` where the additional implementation of `GetLogVal()` really improved the stability of the fit.

alexpearce:
I hadn't heard of `getLogVal` before. [`RooGaussian`](https://root.cern.ch/root/html600/src/RooGaussian.cxx.html) just calls `RooAbsPdf::getLogVal` (which is odd because I guess that would be called anyway…), so seeing as things seem to work at the moment I'll leave it alone.

dominikmuller:
OK sure. I only had the problem once when fitting a narrow gaussian into a wide range that the tails got too small e-300 and numerically you took the log of 0.

Issue #83, opened 2015-02-16
DaVinci Bug - isTheMotherCorrectlyIdentified


chrisburr:
When running over some MC datasets [`TupleToolMCTruth`](https://github.com/alexpearce/CharmProduction/blob/isTheMother-bug-example/options/helpers/tuple_templates.py#L49) and [`TupleToolMCBackgroundInfo`](https://github.com/alexpearce/CharmProduction/blob/isTheMother-bug-example/options/helpers/tuple_templates.py#L100) result in DaVinci crashing.

I've pushed [a branch](https://github.com/alexpearce/CharmProduction/tree/isTheMother-bug-example) which demonstrates the crash in interactive mode when you run `ganga ganga_simulation_S20.py -t MagDown 2015 30000000`. There is a `CRASH` flag in `helpers/tuple_templates.py` which can be switched to enable/disable the problematic tools.

alexpearce:
[Link to mailing list discussion](https://groups.cern.ch/group/lhcb-davinci/Lists/Archive/Flat.aspx?RootFolder=%2Fgroup%2Flhcb-davinci%2FLists%2FArchive%2FSegmentation%20fault%20when%20requiring%20TupleToolMCTruth&FolderCTID=0x01200200B795ECDBD8943F4E90E5F02A017196C9), for reference.

chrisburr:
And to the [JIRA task](https://its.cern.ch/jira/browse/LHCBPS-1400).

alexpearce:
Vava has [committed a fix](https://svnweb.cern.ch/trac/lhcb/changeset/184013) for the background category tool, does this solve things, can we use all the tools OK?

Issue #59, opened 2015-02-05
PropertimeFitter is deprecated


alexpearce:
When creating ntuples, I noticed this in the logs:

```
ToolSvc.properTimeWARNING PropertimeFitter is no longer maintained and thus depreciated.
```

This is because we've got lines like this:

```python
mother_loki_vars = {
    # ...
    'Loki_BPVLTIME': "BPVLTIME('PropertimeFitter/properTime:PUBLIC')",
}
```

Instead we should just be using `BPVLTIME()`. This isn't a big problem for exsiting ntuples, but for new ones we should make sure to switch. (The stripping lines, see #53, do use `BPVLTIME()`, but for the MVA training we use this deprecated fitter, and so also use it in the MVA cut.)

chrisburr:
We can fix this when we get stripping 21, we also need to add `DOCAMIN` as Vava is using it in the HLT selection.

alexpearce:
Ah yes I forgot we'll reprocess for S21. Good plan.

chrisburr:
How can we get `DOCAMIN`? I can [see](https://twiki.cern.ch/twiki/bin/view/LHCb/LoKiParticleFunctions#MINCLAPP_C_type_LoKi_Particles_M) some [possibilities](https://twiki.cern.ch/twiki/bin/view/LHCb/LoKiParticleFunctions#AMAXDOCA_AMINDOCA_Distance_of_cl) but I can't get them to work.

alexpearce:
Have tried just using the functor `DOCAMIN`?

chrisburr:
Yeah [it doesn't exist.](http://lhcb-release-area.web.cern.ch/LHCb-release-area/DOC/davinci/releases/latest/doxygen/d7/dae/namespace_lo_ki_1_1_cuts.html)

alexpearce:
[This looks helpful](https://groups.cern.ch/group/lhcb-davinci/Lists/Archive/Flat.aspx?RootFolder=%2fgroup%2flhcb-davinci%2fLists%2fArchive%2fTupling%20AMAXDOCA&FolderCTID=0x01200200B795ECDBD8943F4E90E5F02A017196C9). During testing, I'd recommend monitoring both `LoKi.Particles.PFunA(AMAXDOCA(''))` and `LoKi.Particles.PFunA(AMINDOCA(''))`, comparing our usual `DOCAMAX` to the new one, so we know this `PFunA` thing is doing stuff correctly.

[...]

PR #86, opened 2015-02-21
Stripping 21

chrisburr:
- [x] Split `decaytreetuples.py`
- [x] Fix proper time is depreciated #59 
- [x] Add `DOCAMIN`
- [ ] Fix `SandboxFile` is depreciated use `LocalFile`
- [ ] Add remaining modes
- [ ] Submit jobs


options/davinci_simulation_2011.py
```diff
@@ -1,3 +1,3 @@
 from helpers import davinci
 
-davinci.configure(year=2011, mc=True, input_type='DST')
+davinci.configure(year=2011, mc=True, input_type='MDST')
```
alexpearce: I expect the simulation will still go to (L)DST, just the centralised stripping output that will go to µDST.

options/ganga_collision_S21.py
```diff
@@ -30,7 +30,7 @@
     exit()
 
 # Determine the stripping and beam energy for the given year
-stripping = ['20', '20r1'][year == '2011']
+stripping = ['21', '21r1'][year == '2011']
```
alexpearce: Is there a Stripping21r1? I think it's just 20 for both years.

options/helpers/tuple_templates.py
```diff
@@ -97,9 +98,9 @@ def decay_tree_tuple(name, decay, mothers, daughters, inputs, mc):
             branch.addTupleTool('TupleToolMCBackgroundInfo')
             # Does the particle ancestry contain a particle with a lifetime
             # above 1e-7 ns? Record the secondaries info if so
-            branch.addTupleTool('TupleToolMCTruth').ToolList = [
-                'MCTupleToolPrompt'
-            ]
+            # branch.addTupleTool('TupleToolMCTruth').ToolList = [
+            #     'MCTupleToolPrompt'
+            # ]
```
alexpearce: This is commented out waiting for a DV release to fix #83?

options/run_stripping_lines_nopids.py
```diff
@@ -78,30 +68,17 @@
...
-        else:
-            # The MB triggers probably aren't running in the 'nominal' TCK used in the MC
-            line_config['HltFilter'] = "HLT_PASS_RE('Hlt1.*')"
-
-        # Make sure there's no pre- or postscales
-        for key in line_config:
-            if key.startswith(('Prescale', 'Postscale')):
-                line_config[key] = 1.0
+        # The MB triggers probably aren't running in the 'nominal' TCK used in the MC
+        line_config['HltFilter'] = "HLT_PASS_RE('Hlt1.*')"
```
alexpearce: I hadn't thought about this before, but this requirement asks for at least one HLT1 line to have fired. Is that really what we want? This filter is applied to the stripping output, which ideally will see NoBias/MinBias stuff, so maybe it's better if this is `None`, like [above](https://github.com/alexpearce/CharmProduction/pull/86/files#diff-0c812ab5daa441aea69f106c214d05b9R73)?

alexpearce:
Deprecated :wink: 

This is looking pretty darn good, thanks for all the work! Splitting up `decaytreetuples.py` makes a lot of sense.

If it's not a massive pain, would it be possible to run over an S20 DST with this branch, and compare the output with running the options in `master` over the same file? It would be a nice sanity check to compare the number of selected entries. (I can see a few things you'd have to change here to make that work, so I'll leave it up to you to decide whether the effort is really worth it, but it has my vote.)

options/davinci_simulation_2011.py
chrisburr: Yeah I changed the simulation ones instead of the collision ones - fixed.

options/ganga_collision_S21.py
chrisburr: Stripping20r1 is a [duplicate of Stripping21 for 2011](https://svnweb.cern.ch/trac/lhcb/browser/Stripping/trunk/Phys/StrippingArchive/python/StrippingArchive/__init__.py#L33).

alexpearce: :+1: derp

options/modes/DsstTophipigamma.py
```diff
@@ -0,0 +1,65 @@
...
+            'Dsst0': '[D*_s+ -> (D_s+ -> (phi(1020) -> K+ K-)) ^gamma]CC',
+            'Ds_pi': '[D*_s+ -> (D_s+ -> (phi(1020) -> K+ K-) ^pi+) gamma]CC',
+            'phi_h1': '[D*_s+ -> (D_s+ -> (phi(1020) -> ^K+ K-) pi+) gamma]CC',
+            'phi_h2': '[D*_s+ -> (D_s+ -> (phi(1020) -> K+ ^K-) pi+) gamma]CC'
+        },
+        inputs_template.format('?????????????'),
+        simulation
+    )
+    # TODO
+    # # Add delta mass
```
chrisburr: Do you know if delta mass is helpful for the photon?

alexpearce: I believe it is. The mass of the D\*s and the Ds are well-defined, so the delta mass is still useful to look at because you should see a peak at the value of the difference of the two (PDG/nominal) masses. The feature you won't see is the ‘threshold’, compared to the usual D*+–D0 delta mass where you see one at the charged pion mass, because the photon is massless (duh!).

There was an [interesting discussion](https://groups.cern.ch/group/lhcb-davinci/Lists/Archive/Flat.aspx?RootFolder=%2Fgroup%2Flhcb-davinci%2FLists%2FArchive%2FCombining%20Gamma%20which%20vertex%20fitters&FolderCTID=0x01200200B795ECDBD8943F4E90E5F02A017196C9) on the mailing list recently about vertexes with photons. Seems that using `DecayTreeFitter` is very helpful. We can use it in a similar way to what we do for the other excited modes, except we probably don't want the primary vertex constraint.

(Actually, I don't think we want it for the other modes either. Ideally, we could have one `DecayTreeFitter` instance with this constraint and one without.)

chrisburr: Am I correct in thinking the primary vertex constraint comes from the `True` part in `DTF_FUN(M, True)`?

options/run_stripping_lines_nopids.py
chrisburr: Done

chrisburr:
I think this is about ready to go now.

A few thoughts:
- Should we be using S20 still for D0->Kpipipi to avoid the [1% prescale](https://svnweb.cern.ch/trac/lhcb/browser/Stripping/trunk/Phys/StrippingSettings/python/StrippingSettings/Stripping21/LineConfigDictionaries_Charm.py?rev=177164#L1609)?
- I've added the `D*0` and `Ds*+` event types to the [twiki page](https://twiki.cern.ch/twiki/bin/view/LHCbPhysics/CharmCrossSection2015#Monte_Carlo)
- I can't find event types for `Dst0_D0pi0_D0gamma,Kpipipi=DecProdCut` or `Dsst_Dsgamma,pipipi=DecProdCut` (but after my last attempt this is probably just me)
- In the bookkeeping I can only find the `Request submitted` event types with 25ns spillover so I think these need to be requested again?
- Once the relevant things are sorted shall I request everything in the table on the twiki?

alexpearce:
1. You're link is to an old revision, the most recent revision shows a [30% prescale](https://svnweb.cern.ch/trac/lhcb/browser/Stripping/trunk/Phys/StrippingSettings/python/StrippingSettings/Stripping21/LineConfigDictionaries_Charm.py?rev=180321#L1408). I think we should stick with all S21. K3Pi is a cross check mode and the sample should still contain enough statistics for us to find any interesting features.
2. Thanks! I agree with the event types you've found for the new modes.
3. I couldn't either. I would be surprised if we really _need_ event types for these modes, as all we're only really interested is the ‘unknown’: the π⁰ and the γ, and we'll get that information from the nominal modes. Still, might be nice to have everything for completeness, just means copying some decay files…
4. Good catch. After talking with my extremely local Charm MC liason, turns out out there are two main “2015” MC models: one for early measurements, at 50 ns bunch spacing with no spillover, and another for nominal 2015 running, at 25 ns bunch spacing with spillover. We want the former, of course, but there was a mixup in the request, so the modes for 2015 conditions that were requested by Dominik were made with the latter. So yes, these two will need resubmitting. We should gather up all the event types we want and do a mammoth request.
5. Yes please, or just let me know as I'm one of the MC liasons now.

chrisburr:
1. Seeming as its 30% I agree.
5. I'll get on it.

alexpearce:
I guess this is waiting on the MC ntuple production?

chrisburr:
Partially, the jobs for the data we already have are currently running.

alexpearce:
OK thanks. Do you have an idea of how long those jobs will take to complete?

chrisburr:
Hopefully in the next couple of days, they're mostly complete I just need to figure out why some jobs have healthy logs but show as failing in Ganga.

alexpearce:
Sounds good, thanks again!

PR #88, opened 2015-03-02
Prettify fit plots

alexpearce:
Clean up the fit plots such that the LaTeX output looks more like what I showed in the [Charm EMTF meeting last week](https://indico.cern.ch/event/365713/contribution/0). Mostly tweaks to make things more legible and a little cleaner.

## ROOT math mode

Commit [`beea9f`](https://github.com/alexpearce/CharmProduction/commit/beea9f5bfe33261f7f4758fd5ae7afbf2cea558c) removes the math mode `$$` wrapping from a lot of strings. This is because ROOT handles TLaTeX in an interesting way (see the commit message).

## Centre axis labels by default

I think it looks better. Thoughts otherwise?

## Sans serif in `.tex` output

Again, I think it looks better. Particularly so in documents where the context is also in sans serif, like presentations, but generally I think it's more readable, mostly because text in figures is generally smaller than the context, and sans serif is more readable at smaller sizes.

## Remove mosaic plots

These weren't really used much anyway, and never look particularly nice. I will shortly be adding a script which recursively saves all `TCanvas` objects in a ROOT file to disk as PDFs, via [`convert_tex_to_pdf`](https://github.com/alexpearce/CharmProduction/blob/90b53c240a61f11fee89d7b7059bb96afa76838f/analysis/charmproduction/plotting/roottex.py#L35). The output of this can be used to make a much nicer looking merged PDF, using something like [PDFjam](http://www2.warwick.ac.uk/fac/sci/statistics/staff/academic-research/firth/software/pdfjam/). Current usage is looking like:

```bash
$ ./plotify.py --recurse --sans-serif ../output/LcTopKpi/2010/MagUp/fitting/fits.root --directory=/Mass
$ pdfnup --no-landscape --suffix nup --nup '5x8' ../output/LcTopKpi/2010/MagUp/fitting/plotify/Mass/c_mass_pT_{7..0}_y_{0..4}.pdf
```


alexpearce:
Example integrated mass plot on 2010 magnet up `DstToD0pi_D0Tohh`.

c_mass_integrated

chrisburr:
First thoughts, that plot looks really nice. Second thoughts, I think the "Candiates / (3MeV/c^2)" label needs shifting to the left a tiny bit.

dominikmuller:
I like the updated style, I have some minor changes in the PIDefficiency branch so I will be careful when finally merging that one.
Concerning the sans serif: putting it in the roottex module will only affect the immediately generated pdf files which is fine, I have no preference about those. Personally, I prefer a uniform font style in my documents so I really like it that the axis labels etc get the matching font (which is still the case if the tex file is included in the document instead of the generated pdf).

alexpearce:
Yeah the sans serif won't affect the `.tex` files. I'm used to using PDFs directly, and I use the standard LaTeX sans serif in my presentations, so when I'm in a rush (always) it's a bit quicker to be able to stick PDFs in, but your right that the `.tex` will blend in with whatever the context is.

We could modify `plotify.py` to output whatever format we want I suppose.

dominikmuller:
Sure, as long as I have the tex files and you in general agree that tiky looks pretty I am happy. I posted something in the issues section which should help speed up latex compilation.

alexpearce:
Yeah I still need to get around to playing with that, does look very useful.

I haven't tested the new stuff on anything other than the output of the fitting, but most of the changes were to fitting-specific code, so everything else _should_ be OK… :four_leaf_clover:

